PR #5, opened 2019-01-08
Import TensorSpec from correct module

wookayin:
As of the most recent stable TensorFlow version (e.g. 1.12.0),
the TensorSpec class is not yet exposed into the outermost tf namespace.

```
>>> tf.TensorSpec
AttributeError: module 'tensorflow' has no attribute 'TensorSpec'
```

To have better compatibility, we can instead import the `TensorSpec`
class from `tensorflow.python.framework.tensor_spec`.

### DIFF
--- a/tf_agents/specs/tensor_spec.py
+++ b/tf_agents/specs/tensor_spec.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


oars (commented):
Thanks for the PR. I'm ironing out our internal tools for merging changes. Once they are fixed I'll merge this. Stay tuned.

PR #32, opened 2019-03-14
Added parallel initialization for ParallelPyEnvironment.

### LINKED ISSUE #31: ParallelPyEnvironment is slow to start
`ParallelPyEnvironment` starts all environments in different processes, but it does so sequentially.

```python
  def start(self):
    logging.info('Starting all processes.')
    for env in self._envs:
      env.start()
    logging.info('All processes started.')
```

```python
  def start(self):
....
    self._process.start()
    result = self._conn.recv()
```

```python
      env = env_constructor()
      action_spec = env.action_spec()
      conn.send(self._READY)  # Ready.
```

If `env_constructor` takes, for example, 10 seconds to finish, then for 64 agents I have to wait 640 seconds. It makes it hard to experiment with heavy environments.

It would be better if we first spawned all start processes, and then waited for all of them to finish.

PeterZhizhin:
As is has already been discussed in #31, ParallelPyEnvironment takes too
much time to initialize for slow-starting environments due to sequential
`env.start()` calls.

This commit fixes #31 and adds tests for the fix.
I have also tested this implementation with a Unity environment. Initialization time dropped from 160 seconds to ~10 seconds for 16 environments.

### DIFF
--- a/tf_agents/environments/parallel_py_environment.py
+++ b/tf_agents/environments/parallel_py_environment.py
@@ -60,6 +60,7 @@ class ParallelPyEnvironment(py_environment.PyEnvironment):
     self._envs = [ProcessPyEnvironment(ctor, flatten=flatten)
                   for ctor in env_constructors]
     self._num_envs = len(env_constructors)
+    self._blocking = blocking
     self.start()
     self._action_spec = self._envs[0].action_spec()
     self._observation_spec = self._envs[0].observation_spec()
@@ -68,13 +69,16 @@ class ParallelPyEnvironment(py_environment.PyEnvironment):
       raise ValueError('All environments must have the same action spec.')
     if any(env.time_step_spec() != self._time_step_spec for env in self._envs):
       raise ValueError('All environments must have the same time_step_spec.')
-    self._blocking = blocking
     self._flatten = flatten
 
   def start(self):
-    logging.info('Starting all processes.')
+    logging.info('Spawning all processes.')
     for env in self._envs:
-      env.start()
+      env.start(wait_to_start=self._blocking)
+    if not self._blocking:
+      logging.info('Waiting for all processes to start.')
+      for env in self._envs:
+          env.wait_start()
     logging.info('All processes started.')
 
   @property
@@ -189,14 +193,23 @@ class ProcessPyEnvironment(object):
     self._action_spec = None
     self._time_step_spec = None
 
-  def start(self):
-    """Start the process."""
+  def start(self, wait_to_start=True):
+    """Start the process.
+
+    Args:
+      wait_to_start: Whether the call should wait for an env initialization.
+    """
     self._conn, conn = multiprocessing.Pipe()
     self._process = multiprocessing.Process(
         target=self._worker,
         args=(conn, self._env_constructor, self._flatten))
     atexit.register(self.close)
     self._process.start()
+    if wait_to_start:
+      self.wait_start()
+
+  def wait_start(self):
+    """Wait for the started process to finish initialization"""
     result = self._conn.recv()
     if isinstance(result, Exception):
       self._conn.close()
--- a/tf_agents/environments/parallel_py_environment.py
+++ b/tf_agents/environments/parallel_py_environment.py
@@ -78,7 +78,7 @@ class ParallelPyEnvironment(py_environment.PyEnvironment):
     if not self._blocking:
       logging.info('Waiting for all processes to start.')
       for env in self._envs:
-          env.wait_start()
+        env.wait_start()
     logging.info('All processes started.')
 
   @property
@@ -209,7 +209,7 @@ class ProcessPyEnvironment(object):
       self.wait_start()
 
   def wait_start(self):
-    """Wait for the started process to finish initialization"""
+    """Wait for the started process to finish initialization."""
     result = self._conn.recv()
     if isinstance(result, Exception):
       self._conn.close()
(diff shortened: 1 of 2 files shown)


PeterZhizhin:
I signed it!

oars:
Thanks! Awesome contribution. I'll keep an eye out for the test as I worry it could end up being slightly flaky with the timing checks. Easy enough to generalize if it does.

PeterZhizhin:
@oars It would be awesome if you found another way of testing this without `time.sleep()` calls. I have tried to find other ways, but I didn't succeed.

tf_agents/environments/parallel_py_environment.py
```diff
@@ -189,14 +193,23 @@ def __init__(self, env_constructor, flatten=False):
...
     self._process = multiprocessing.Process(
         target=self._worker,
         args=(conn, self._env_constructor, self._flatten))
     atexit.register(self.close)
     self._process.start()
+    if wait_to_start:
+      self.wait_start()
+
+  def wait_start(self):
+    """Wait for the started process to finish initialization"""
```
oars: Nit: 

```suggestion
    """Wait for the started process to finish initialization."""
```

tf_agents/environments/parallel_py_environment.py
```diff
@@ -68,13 +69,16 @@ def __init__(self, env_constructors, blocking=False, flatten=False):
...
   def start(self):
-    logging.info('Starting all processes.')
+    logging.info('Spawning all processes.')
     for env in self._envs:
-      env.start()
+      env.start(wait_to_start=self._blocking)
+    if not self._blocking:
+      logging.info('Waiting for all processes to start.')
+      for env in self._envs:
+          env.wait_start()
```
oars:
```suggestion
        env.wait_start()
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -21,6 +21,7 @@
 
 import collections
 import functools
```
oars, line 23: `import multiprocessing.dummy as multiprocessing`

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -31,19 +32,29 @@
 from tf_agents.specs import array_spec
 
 
+class SlowStartingEnvironment(random_py_environment.RandomPyEnvironment):
+  def __init__(self, *args, time_sleep=0.1, **kwargs):
```
oars: We need python 2 support.

```suggestion
  def __init__(self, *args, **kwargs):
    time_sleep = kwargs.pop('time_sleep', 1.0)
     super().__init__(*args, **kwargs)
￼     time.sleep(time_sleep)
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -31,19 +32,29 @@
...
+    super().__init__(*args, **kwargs)
+    time.sleep(time_sleep)
+
+
 class ParallelPyEnvironmentTest(tf.test.TestCase):
```
oars, line 44: Threads seem to be more reliable for timing than processes.

```
  def setUp(self):
    parallel_py_environment.multiprocessing = multiprocessing
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -80,6 +91,36 @@ def test_step(self):
                         time_step2.observation.shape)
     env.close()
 
+  def test_non_blocking_start_processes_in_parallel(self):
+    self._set_default_specs()
+    constructor = functools.partial(SlowStartingEnvironment,
+                                    self.observation_spec,
+                                    self.action_spec, time_sleep=0.1)
```
oars:
```suggestion
                                    self.action_spec, time_sleep=1.0)
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -80,6 +91,36 @@ def test_step(self):
...
+    start_time = time.time()
+    self._make_parallel_py_environment(constructor=constructor,
+                                       num_envs=10,
+                                       blocking=False)
+    end_time = time.time()
+    self.assertLessEqual(end_time - start_time, 0.5,
```
oars:
```suggestion
    self.assertLessEqual(end_time - start_time, 5.0,
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -80,6 +91,36 @@ def test_step(self):
...
+                         msg=('Expected all processes to start together, '
+                              'got {} wait time').format(
+                           end_time - start_time))
+
+  def test_blocking_start_processes_one_after_another(self):
+    self._set_default_specs()
+    constructor = functools.partial(SlowStartingEnvironment,
+                                    self.observation_spec,
+                                    self.action_spec, time_sleep=0.05)
```
oars:
```suggestion
                                    self.action_spec, time_sleep=1.0)
```

tf_agents/environments/parallel_py_environment_test.py
```diff
@@ -80,6 +91,36 @@ def test_step(self):
...
+    start_time = time.time()
+    self._make_parallel_py_environment(constructor=constructor,
+                                       num_envs=10,
+                                       blocking=True)
+    end_time = time.time()
+    self.assertGreater(end_time - start_time, 0.3,
```
oars:
```suggestion
    self.assertGreater(end_time - start_time, 10,
```

oars (commented):
For now let's use threading instead of processes (with this 1000/1000 runs passed for me, as opposed to 990/1000 with multi process.

Added some nits on formatting and python 2 compatibility.

PeterZhizhin:
Thank you for the review. I have resolved all your comments. Also added close() call for added tests, otherwise the test just hangs (atexit is not valid for threads).

PR #33, opened 2019-03-14
Added missing Gym spaces

### LINKED ISSUE #28: MultiDiscrete spaces support
Hello. Thank you for an amazing product, I am experimenting with this right now.

I would like to know, why `MultiDiscrete` spaces for Gym wrappers are not supported. I don't see any reasons why it should be difficult to support the space. It should be similar to a tuple of discrete spaces which you already support.

Could you please provide some clarification?

PeterZhizhin:
As discussed in #28, tf-agents lack MultiDiscrete space support.

This commit fixes #28 by adding the space as well as MultiBinary space.

Also added tests for the spaces.

### DIFF
--- a/tf_agents/environments/gym_wrapper_test.py
+++ b/tf_agents/environments/gym_wrapper_test.py
@@ -46,7 +46,7 @@ class GymWrapperSpecTest(absltest.TestCase):
     spec = gym_wrapper._spec_from_gym_space(multi_discrete_space)
 
     self.assertEqual((4,), spec.shape)
-    self.assertEqual(np.uint32, spec.dtype)
+    self.assertEqual(np.int32, spec.dtype)
     np.testing.assert_array_equal(
       np.array([0, 0, 0, 0], dtype=np.int),
       spec.minimum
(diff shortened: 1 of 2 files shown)


PeterZhizhin:
Looks like the internal tests from copybara are hanging. Could you please have a look?

PeterZhizhin:
I have changed the default MultiDiscrete type from uint32 to int32. Tensorflow does not support some uint32 ops.

tf_agents/environments/gym_wrapper_test.py
```diff
@@ -41,6 +41,36 @@ def test_spec_from_gym_space_discrete(self):
     self.assertEqual(0, spec.minimum)
     self.assertEqual(2, spec.maximum)
 
+  def test_spec_from_gym_space_multi_discrete(self):
+    multi_discrete_space = gym.spaces.MultiDiscrete([1, 2, 3, 4])
+    spec = gym_wrapper._spec_from_gym_space(multi_discrete_space)
+
+    self.assertEqual((4,), spec.shape)
+    self.assertEqual(np.uint32, spec.dtype)
```
oars: Need to update the test as well.

PR #35, opened 2019-03-15
Fixed crashes when debug_summaries is true in PPO

### LINKED ISSUE #29: debug_summaries makes PPOAgent fail when observation is a tuple
When an agent observation is a tuple and `debug_summaries` is set to True, the agent crashes in a `tf.compat.v2.summary.histogram` call.

Let's have a look at the code that creates the problem:
```python
    observation = time_steps.observation
    if debug_summaries:
      tf.compat.v2.summary.histogram(
          name='observations', data=observation, step=self.train_step_counter)
```

If `observation` is a tuple, the call crashes.

Probably before fixing this, we should identify other possible instances of this pattern to fix it everywhere.

PeterZhizhin:
As discussed in #29, nested observations crash the agent with
debug_summaries set to True.

While I was fixing the bug, I found that nested actions also cause the
agent to crash in multiple places. I have also found another bug in
ppo_agent_test.py that caused another crash, fixed this as well.

The commit fixes #29.


tf_agents/agents/ppo/ppo_agent_test.py
```diff
@@ -69,7 +68,13 @@ def _get_normal_distribution_spec(self, sample_spec):
 
   def call(self, inputs, unused_step_type=None, network_state=()):
     hidden_state = tf.cast(tf.nest.flatten(inputs), tf.float32)[0]
-    batch_squash = network_utils.BatchSquash(self._outer_rank)
+
+    # Calls coming from agent.train() has a time dimension. Direct loss calls
+    # may not have a time dimension. It order to make BatchSquash work, we need
+    # to specify the outer dimension properly.
+    has_time_dim = hidden_state.ndim == 3
```
oars: Need tf 1.0 compatibility: 

```suggestion
    has_time_dim = tf.rank(hidden_state) == 3
```

oars:
Tests seem to be failing, please make sure they are passing.

PeterZhizhin:
Tests pass locally for me. Could you please provide more information?

PeterZhizhin:
Ok. I had to create a separate test environment with TF 1.14.
Do you know whether there is an option to run tests for pull requests in GitHub as a check, similarly to CLA and Copybara checks?

oars:
I'll be setting up Travis for this repo soon which should help run on the variety of targets we need to test.

oars:
Also, thanks for all the contributions!

Issue #30, opened 2019-03-13
Training PPO in non-episodic environments


PeterZhizhin:
After reviewing PPOAgent code, I have found the following piece of code:
```python
    valid_mask = ppo_utils.make_timestep_mask(next_time_steps)

    if weights is None:
      weights = valid_mask
    else:
      weights *= valid_mask

....

    value_preds, unused_policy_state = self._collect_policy.apply_value_network(
        time_steps.observation, time_steps.step_type, policy_state=policy_state)
    value_estimation_error = tf.math.squared_difference(returns, value_preds)
    value_estimation_error *= weights
```

This means that if an agent runs in a non-episodic environment (the environment never resets and can be run infinitely), the agent will never learn anything because `weights` will be always zero.

The agent can get some reward, but due to weight being zero, training will not happen.

This also makes it harder to train the agent with `DynamicStepDriver`. Only latest steps before reset will be used for training.

Axel-CH:
Hi @PeterZhizzhin, thank you for your PR, I have the same issue. Could you add an exemple (gym or other env) of training with Training PPO in non-episodic environment ?

DanielWicz:
> Hi @PeterZhizzhin, thank you for your PR, I have the same issue. Could you add an exemple (gym or other env) of training with Training PPO in non-episodic environment ?

Flatlat (http://flatland-rl-docs.s3-website.eu-central-1.amazonaws.com/index.html) is such an example. Nevertheless, there's a solution to this issue ?

tfboyd:
Little cleanup and issue bankruptcy. Issue is ~2 years old. I am sorry that this did not end with a solution. If you find this issue in the future please open a new one and reference this one if desired.

PR #36, opened 2019-03-15
Added an option for PPO to train on incomplete episodes.

PeterZhizhin:
As discussed in #30, right now PPO discards all rewards that were
found on not completed episodes. This commit addresses this issue and
adds an option to train on episodes that were not completed.

Also added tests to ensure that loss is not zero when training on
incomplete episodes depending on the flags.


tf_agents/agents/ppo/ppo_agent.py
```diff
@@ -118,6 +118,7 @@ def __init__(self,
                normalize_rewards=True,
                reward_norm_clipping=10.0,
                normalize_observations=True,
+               train_only_on_full_episodes=True,
```
ebrevdo: @oars should we make the default false here?

tf_agents/agents/ppo/ppo_agent.py
```diff
@@ -487,7 +492,10 @@ def _train(self, experience, weights):
         experience.observation, experience.step_type, policy_state=policy_state)
     value_preds = tf.stop_gradient(value_preds)
 
-    valid_mask = ppo_utils.make_timestep_mask(next_time_steps)
+    if self._train_only_on_full_episodes:
+      valid_mask = ppo_utils.make_timestep_mask(next_time_steps)
+    else:
+      valid_mask = tf.ones(next_time_steps.step_type.shape)
```
ebrevdo: is there a way to make valid_mask None here and respect that throughout the rest of the code, thereby avoiding an extra useless tensor?

PeterZhizhin:
I have resolved the issue with a potentially unused tensor. Waiting for @oars to answer.

tf_agents/agents/ppo/ppo_agent.py
PeterZhizhin: I would also like to know whether there is a reason behind this `valid_mask`?
I didn't see this implemented in any common RL libraries such as OpenAI Baselines or Ray RLlib.

oars: The mask is needed because of the way data is added to the RB. In general observers are not filtering transitions between `step_type.LAST` and `step_type.FIRST`. I also don't think all agents are handling this correctly, in general we need to fix this.

@ebrevdo defaulting to False would definetly be the safest option to make sure behavior doesn't change for users.

@sguada Know if any other parts of PPO rely on the experience being full episodes?

sguada: I think what we need is to bootstrap the end of episodes using the value function if it is not an end of episode.

But we still need to use the mask to avoid invalid transitions between `LAST` and `FIRST`.

Rename to `train_only_on_partial_episodes=True,`

tf_agents/agents/ppo/ppo_agent.py
```diff
@@ -487,11 +492,13 @@ def _train(self, experience, weights):
         experience.observation, experience.step_type, policy_state=policy_state)
     value_preds = tf.stop_gradient(value_preds)
 
-    valid_mask = ppo_utils.make_timestep_mask(next_time_steps)
```
sguada, line 495: pass `train_only_on_partial_episodes` to `make_timestep_mask` and build an appropriate mask.

keep `valid_mask` in the rest of code.

sguada (commented):
```
def make_timestep_mask(batched_next_time_step, partial_episodes=False):
  """Create a mask for final incomplete episodes and episode transitions.
  Args:
    batched_next_time_step: Next timestep, doubly-batched
      [batch_dim, time_dim, ...].
    partial_episodes: Bool, when True allow partial episodes.
  Returns:
    A mask, type tf.float32, that is 0.0 for all between-episode timesteps
      (batched_next_time_step is FIRST), or where the episode is
      not compelete, so the return computation would not be correct.
  """
  # 1.0 for timesteps of all complete episodes. 0.0 for incomplete episode at
  #   the end of the sequence.
  if partial_episodes:
   episode_is_complete = tf.ones_like(batched_next_time_step.is_last(), dtype=tf.bool)
  else:
   episode_is_complete = tf.cumsum(
      tf.cast(batched_next_time_step.is_last(), tf.float32),
      axis=1,
      reverse=True) > 0

  # 1.0 for all valid timesteps. 0.0 where between episodes.
  not_between_episodes = ~batched_next_time_step.is_first()

  return tf.cast(episode_is_complete & not_between_episodes, tf.float32)
```

tf_agents/agents/ppo/ppo_agent.py
PeterZhizhin: Used `train_on_partial_episodes` instead of `train_only_on_partial_episodes` as it should also work in episodic envs as well.

Bootstrapped the reward using the value function as well.

tf_agents/agents/ppo/ppo_agent.py
PeterZhizhin: New commits solve the described issues.

PeterZhizhin:
I see that there is an error while migrating the change. Could you please clarify what is the problem?

sguada:
You probably need to sync your branch to the latest code and merge any conflicts.

andreas-repp:
@PeterZhizhin: Can you update your branch so they can merge it? I'd really like to see this implemented. Thanks already for providing this PR in the first place!

tfboyd:
@sguada @ebrevdo   The conflicts are not that hard to resolve.  Assuming the person does not resond (it has been a long time); I am happy to fix the conflicts and finish this off if this is something we want.  It looks like it was almost done.

tf_agents/agents/ppo/ppo_agent.py
```diff
@@ -164,6 +165,9 @@ def __init__(self,
       reward_norm_clipping: Value above an below to clip normalized reward.
       normalize_observations: If true, keeps moving mean and variance of
         observations and normalizes incoming observations.
+      train_on_partial_episodes: If True (default False), then data from
```
ebrevdo, line 168: i wonder if we can just always train on partial episodes.  @oars wdyt?  can we get rid of this arg and always assume true?

ebrevdo:
This looks great; I just want to see if we can always train from partial episodes.  If so we can get rid of one of the init args.

tf_agents/agents/ppo/ppo_agent.py
oars: I think it should be fine. Previous use cases will still work as they are feeding full episodes already.

tfboyd:
I am 95% sure the copybara process is failing because of the branch conflict.  The PR branch does not have /docs, which is expected by copybara, which would be fine but I suspect because it is not merging due to the conflict the branch copybara is seeing does not have /docs.

tfboyd:
I am making a new PR to fix the conflict and preserve the commits and credit for the original author.  Should be done in a few minutes....fingers crossed.

tfboyd:
Already I merged Peter's commits and made a new PR fixing the conflicts.  I am running the unit tests as well.  Please go here.  https://github.com/tensorflow/agents/pull/306

PR #46, opened 2019-03-20
Added missing check to PyEnvironment.step(action)

jmribeiro:
Added missing check to PyEnvironment.step(action) showcase, where it wasn't checking if the current_time_step was None (and resetting if so).


tf_agents/colabs/2_environments_tutorial.ipynb
```diff
@@ -155,6 +155,8 @@
         "\n",
         "  def step(self, action):\n",
         "    \"\"\"Apply action and return new time_step.\"\"\"\n",
+        "    if self._current_time_step is None:",
```
kbanoop: missing \n at the end of both lines

jmribeiro: Corrected, thanks!

PR #58, opened 2019-04-02
Fix tf-nightly typo in README

seungjaeryanlee:
The PyPI package `tf-nightly` was written as `tf-nighly` in one occasion.


seungjaeryanlee:
I signed it!

PR #60, opened 2019-04-03
Fix a bug in the __eq__ method for BoundedArraySpec.

npfp:
When one.minimum property is an empty array
(for example in a child class)
and other.minimum contains a single element
the equality used to be true.
(see for example https://stackoverflow.com/a/10580782/4282745).

### DIFF
--- a/tf_agents/specs/array_spec_test.py
+++ b/tf_agents/specs/array_spec_test.py
@@ -341,14 +341,14 @@ class BoundedArraySpecTest(parameterized.TestCase):
     self.assertNotEqual(spec_1, spec_2)
 
   def testNotEqualEmptyMinimum(self):
-      class MockBoundedArraySpec(array_spec.BoundedArraySpec):
-        @property
-        def minimum(self):
-          return np.array([])
+    class MockBoundedArraySpec(array_spec.BoundedArraySpec):
+      @property
+      def minimum(self):
+        return np.array([])
 
-      spec_1 = array_spec.BoundedArraySpec( (1,), np.int32, minimum=[-1.6], maximum=[1.0])
-      spec_2 = MockBoundedArraySpec( (1,), np.int32, minimum=[0.0], maximum=[1.0])
-      self.assertNotEqual(spec_1, spec_2)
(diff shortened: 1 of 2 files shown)


tf_agents/specs/array_spec_test.py
```diff
@@ -340,6 +340,16 @@ def testNotEqualDifferentMinimum(self):
         (1, 2), np.int32, minimum=[0.0, 0.0], maximum=[1.0, 1.0])
     self.assertNotEqual(spec_1, spec_2)
 
+  def testNotEqualEmptyMinimum(self):
+    class MockBoundedArraySpec(array_spec.BoundedArraySpec):
+      @property
+      def minimum(self):
+        return np.array([])
+
+    spec_1 = array_spec.BoundedArraySpec( (1,), np.int32, minimum=[-1.6], maximum=[1.0])
```
egonina, line 349: nit: remove the space before (1,) in BoundedArraySpec

tf_agents/specs/array_spec_test.py
```diff
@@ -340,6 +340,16 @@ def testNotEqualDifferentMinimum(self):
         (1, 2), np.int32, minimum=[0.0, 0.0], maximum=[1.0, 1.0])
     self.assertNotEqual(spec_1, spec_2)
 
+  def testNotEqualEmptyMinimum(self):
```
egonina, line 343: can you cadd a test for equal on two instances of MockBoundedArraySpec? (i.e. empty is equal emtpy)

egonina (commented):
Thanks for the pull request!

tf_agents/specs/array_spec_test.py
npfp: Thx for the comment.

I've just figured out that my MockBoundedArraySpec breaks the contract specified in the `__init__` method:

```python
      minimum: Number or sequence specifying the maximum element bounds
        (inclusive). Must be broadcastable to `shape`.
```

So the code was fine as it was (shapes are the same as checked by the call to super().__eq__ (and so minimums are broadcastable to she same shape)). 

I'll fix my own child classes and revert my PR.

Apologies for the waste of time.

egonina: no problem!

PR #68, opened 2019-04-07
Dynamic array spec

npfp:
Make (Bounded)ArraySpec allow for dynamic shape following the same convention than in tensorflow: any None dimension represents a dynamic dimension. This is to allow for the cases where observations can have a dynamic dimension: for example if the observation can be empty or a concatenation of several atomic observations.

Check that the conversion to tensor works fine.

Implement an example of RandomPyEnvironment where the observation can have dynamic dimensions drown between 0 and 4.

Add a helper function to set the unknown dimensions of a dynamic shape to given values.

### DIFF
--- a/tf_agents/environments/random_py_environment.py
+++ b/tf_agents/environments/random_py_environment.py
@@ -169,17 +169,3 @@ class RandomPyEnvironment(py_environment.PyEnvironment):
           format(mode))
 
     return self._rng.randint(0, 256, size=self._render_size, dtype=np.uint8)
-
-
-class DynamicRandomPyEnvironment(RandomPyEnvironment):
-
-  def _get_observation(self):
-    # Observation has a dynamic shape sampled uniformly between 0 and 4.
-    num_dynamic_dims = len([d for d in self._observation_spec.shape if d is None])
-    observation_spec = self._observation_spec
-    if num_dynamic_dims > 0:
-        dynamic_dims = self._rng.randint(low=-1, high=5, size=num_dynamic_dims)
-        observation_spec = array_spec.set_dynamic_dims_nest(observation_spec, dynamic_dims)
-    batch_size = (self._batch_size,) if self._batch_size else ()
-    return array_spec.sample_spec_nest(observation_spec, self._rng, batch_size)
-
--- a/tf_agents/environments/random_py_environment.py
+++ b/tf_agents/environments/random_py_environment.py
@@ -174,11 +174,11 @@ class RandomPyEnvironment(py_environment.PyEnvironment):
 class DynamicRandomPyEnvironment(RandomPyEnvironment):
 
   def _get_observation(self):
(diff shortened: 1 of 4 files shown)


tf_agents/environments/random_py_environment.py
```diff
@@ -169,3 +169,17 @@ def render(self, mode='rgb_array'):
           format(mode))
 
     return self._rng.randint(0, 256, size=self._render_size, dtype=np.uint8)
+
+
+class DynamicRandomPyEnvironment(RandomPyEnvironment):
```
sguada: This class doesn't make much sense as an environment, what is suppose to do? If you need this for the test, then move it to the test file or to utils_test.py

tf_agents/environments/random_py_environment.py
```diff
@@ -169,3 +169,17 @@ def render(self, mode='rgb_array'):
...
+
+  def _get_observation(self):
+    # Observation has a dynamic shape sampled uniformly between 0 and 4.
```
sguada: We cannot hard-code that the shape is sample uniformly between 0 and 4. You should add either a property to set that or pass an extra argument to __init__ to decide what to do in that case.

sguada (commented):
It is not very clear the usage of this PR. Also they way Nones are handled seems brittle.

tf_agents/environments/random_py_environment.py
npfp: It was indeed only designed as an example of use case as well as for the case. I'll move it to the test file.

tf_agents/environments/random_py_environment.py
npfp: I'll make the change.

npfp:
The purpose is to allow cases where observation have dynamic shapes: for example if there are random delays in your environment, you might receive no observation for a while and then a bunch of observations.

sguada:
Specs represent the dimension of the observation without including time or batch dimensions. So if you just want to batch multiple observations then you don't to specify None for the batch_dimension.

npfp:
When I said multiple observations, I had in mind multiple observations for a single agent, not a batch of observations for a batch of agents.

Let's say the agent plays a lagged version of Starcraft : a piece of observation is a screenshot (tensor with shape (n, m) for example). Some times the agent doesn't get any screenshot, sometime he receives a few of them at the same time. So the observation is a list of screenshots with variable length.

Our proposal is to implement it using a spec of (dynamic) shape `(None, n, m)` because we'd go this way in tensorflow. I'd be happy to know another ways to implement the previous example. We considered not using a dynamic batch but a list of pieces of observation but it didn't work either as you'd need to specify a fixed length fort the list.

tfboyd:
Last comment was from April-2019.  This has become stale.  I am sorry we were not able to find a path to merge the PR.  The team is focusing on new PRs and working to bring PRs to a merge or close state in a more timely manner.  Thank you for taking the time to create the PR and have a productive back and forth discussion.

Issue #67, opened 2019-04-06
py_driver constructor, question  


idigitopia:
max_steps = max_steps or 0 
    max_episodes = max_episodes or 0
    if max_steps < 1 and max_episodes < 1:
      raise ValueError(
          'Either "max_steps" or "max_episodes" should be greater than 0.')
    super(PyDriver, self).__init__(env, policy, observers)
    self._max_steps = max_steps or np.inf
    self._max_episodes = max_episodes or np.inf 

is there a scenario where  the max_stes will be assigned np.inf ? 
I could not understand the logic behind it. 

Sorry if this is too basic. :)

sguada:
So the logic behind it is two fold:
- You need to specify at least one parameter `max_steps` or `max_episodes` so if neither of them are specified then it would raise an error.
- If only one is specified then then the other would be converted to np.inf (so it won't be a limit)

Maybe the following boolean examples help:
```
None or 0 => 0
0 or np.inf => np.inf
```

A different way to write the code would be:
```
if max_steps is None and max_episodes is None:
 raise ValueError(
'Either "max_steps" or "max_episodes" should be greater than 0.')
if max_steps is None:
  max_steps = np.inf
if max_episodes is None:
  max_episodes = np.inf
self._max_steps = max_steps
self._max_episodes = max_episodes
```

sguada:
Re-open if needed.

PR #74, opened 2019-04-11
Updated PyDriver Constructor to be more readable.

idigitopia:
Resolves Issue #67 

```
import numpy as np
import itertools

# Very Confusing
def old_func(max_steps, max_episodes):
	max_steps = max_steps or 0
	max_episodes = max_episodes or 0

	if max_steps < 1 and max_episodes < 1:
		return 'Error'

	_max_steps = max_steps or np.inf
	_max_episodes = max_episodes or np.inf
	return (_max_steps, _max_episodes)


# More Readable
def new_func(max_steps, max_episodes):
	if max_steps in (None, 0) and max_episodes in (None, 0):
		return 'Error'

	_max_steps = max_steps or np.inf
	_max_episodes = max_episodes or np.inf
	return (_max_steps, _max_episodes)


test_cases = [i for i in itertools.product([None, 0, 4], repeat=2)]

for case in test_cases:
	assert old_func(*case) == new_func(*case)
```

### DIFF
--- a/tf_agents/drivers/py_driver.py
+++ b/tf_agents/drivers/py_driver.py
@@ -50,7 +50,7 @@ class PyDriver(driver.Driver):
     Raises:
       ValueError: If both max_steps and max_episodes are None.
     """
-    if max_steps in (None, 0) and max_episodes in (None, 0):
+    if (max_steps is None or max_steps <1) and (max_episodes is None or max_episodes <1):
       raise ValueError(
           'Either `max_steps` or `max_episodes` should be greater than 0.')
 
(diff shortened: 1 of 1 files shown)


idigitopia:
I signed it!

tf_agents/drivers/py_driver.py
```diff
@@ -50,9 +50,7 @@ def __init__(self,
     Raises:
       ValueError: If both max_steps and max_episodes are None.
     """
-    max_steps = max_steps or 0
-    max_episodes = max_episodes or 0
-    if max_steps < 1 and max_episodes < 1:
+    if max_steps in (None, 0) and max_episodes in (None, 0):
```
sguada: This doesn't test that max_steps or max_episodes is not negative.

sguada (changes requested):

tfboyd:
Last activity was May 2019.  A lot of time has passed, closing the issue due to lack of activity.

PR #115, opened 2019-05-22
Add __version__, fixes #114

### LINKED ISSUE #114: The version number is missing (PEP 396)
There's currently no way to check which version of TF-Agents is installed. I know there is no stable version available yet, but this should not prevent having a version we can refer to. It will have to be done at one point, so why not now?

Current behavior:

```pycon
>>> import tf_agents
>>> tf_agents.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'tf_agents' has no attribute '__version__'
```

Desired behavior:

```pycon
>>> import tf_agents
>>> tf_agents.__version__
'0.1.0-dev20190522'
```

ageron:
Just imports `__version__` from `tf_agents.version` for PEP 396.

### DIFF
--- a/tf_agents/__init__.py
+++ b/tf_agents/__init__.py
@@ -64,6 +64,7 @@ def _ensure_tf_install():  # pylint: disable=g-statement-before-imports
 
 _ensure_tf_install()
 
+from tf_agents.version import __version__
 
 # Cleanup symbols to avoid polluting namespace.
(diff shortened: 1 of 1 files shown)


egonina:
Thank you for the PR!

ageron:
Thanks for merging! 👍

PR #116, opened 2019-05-23
If the current time step is the last, step() should call reset() automatically

ageron:
Most of the time, you want to reset the environment after an episode is ended, so why not do it automatically when `step()` is called? I made this change and updated the environments notebook as well.


egonina:
Thank you for the PR! CC @kbanoop - was there a reason we wanted to have env.reset() called explicitly?

tf_agents/environments/py_environment.py
```diff
@@ -168,7 +168,7 @@ def step(self, action):
         observation: A NumPy array, or a nested dict, list or tuple of arrays
           corresponding to `observation_spec()`.
     """
-    if self._current_time_step is None:
+    if self._current_time_step is None or self._current_time_step.is_last():
```
kbanoop, line 171: One challenge here is that python environments could be batched. So each call to step would be stepping each of the sub environments, and it is possible that only some of them have current_time_step.is_last() == True.

ageron: Ah yes, I didn't think about batched environments, thanks. Mmh, not sure what's the best option here: we could reset if the environment is not batched, which would probably take care of batched environments as well, indirectly (since they are probably made of multiple non batched environments, which will get a call to `step()` and reset if needed. However, I'm not sure what the implications are.

kbanoop: Good point, yeah I think that should handle the batched environments. 

If we go ahead with this change, we should also make similar changes in the TF environment so that things are consistent, 
https://github.com/tensorflow/agents/blob/master/tf_agents/environments/tf_environment.py

And also in the different suites. Maybe just this file I think:
https://github.com/tensorflow/agents/blob/58c7a5fdb01359ffb2cefcda9350aa65a0907eb5/tf_agents/environments/gym_wrapper.py#L153

tfboyd:
Closing,  we worked hard to catch up on some of the really old PRs, but this one may have to be a casualty.  We have gotten much better at working the PRs faster.

PR #118, opened 2019-05-24
Use tf.image.resize() rather than cv2, and extend gym.core.Wrapper

ageron:
TensorFlow has a `tf.image.resize()` method, why not use it instead of have a dependency on `cv2`?

Also, the `AtariPreprocessing` class does not have a `close()` method, and `render()` should have `mode='human'` by default, like other Gym environments. This can be solved easily by inheriting from the `gym.core.Wrapper` class, like all other wrappers.

### DIFF
--- a/tf_agents/environments/atari_preprocessing.py
+++ b/tf_agents/environments/atari_preprocessing.py
@@ -30,14 +30,15 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
+from gym.core import Wrapper
 from gym.spaces.box import Box
 import numpy as np
 import gin.tf
-import cv2
+import tensorflow as tf
 
 
 @gin.configurable
-class AtariPreprocessing(object):
+class AtariPreprocessing(Wrapper):
   """A class implementing image preprocessing for Atari 2600 agents.
 
   Specifically, this provides the following subset from the JAIR paper
@@ -54,7 +55,7 @@ class AtariPreprocessing(object):
   """
 
   def __init__(self,
-               environment,
+               env,
                frame_skip=4,
                terminal_on_life_loss=False,
                screen_size=84):
@@ -70,6 +71,16 @@ class AtariPreprocessing(object):
     Raises:
       ValueError: if frame_skip or screen_size are not strictly positive.
     """
+    super(AtariPreprocessing, self).__init__(env)
+
+    # Return the observation space adjusted to match the shape of the processed
+    # observations.
+    self.observation_space = Box(
+        low=0,
+        high=255,
+        shape=(screen_size, screen_size, 1),
+        dtype=np.uint8)
+
     if frame_skip <= 0:
       raise ValueError(
           'Frame skip should be strictly positive, got {}'.format(frame_skip))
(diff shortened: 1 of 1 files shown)


ageron:
Note: I timed `tf.image.resize()` and `cv2.resize()`, and it seems that `cv2.resize()` is about 30% faster than `tf.image.resize()`. Probably not a big deal since we're talking about a difference of about 100μs, but still worth mentioning. If we train a DQN for 50 million steps, this difference adds up to a total of about 2 hours on my computer, but since the GPU will probably be the bottleneck, rather than the CPU, it probably does not matter.
Perhaps one solution would be to check whether cv2 is installed or not, and use it if it is, otherwise use TensorFlow. However, although their outputs are very close, they are not perfectly identical, so it might be weird to have different results depending on which libraries are installed.

tf_agents/environments/atari_preprocessing.py
```diff
@@ -31,13 +31,14 @@
 from __future__ import print_function
 
 import gin
+from gym.core import Wrapper
 from gym.spaces.box import Box
 import numpy as np
-import cv2
+import tensorflow as tf
```
sguada: Not sure if it is better to depend on TF than on cv2 here?

tf_agents/environments/atari_preprocessing.py
```diff
@@ -224,8 +195,8 @@ def _pool_and_resize(self):
...
-    transformed_image = cv2.resize(
-        self.screen_buffer[0], (self.screen_size, self.screen_size),
-        interpolation=cv2.INTER_AREA)
-    int_image = np.asarray(transformed_image, dtype=np.uint8)
-    return np.expand_dims(int_image, axis=2)
+    transformed_image = tf.image.resize(
+        images=self.screen_buffer[0][..., np.newaxis],
+        size=(self.screen_size, self.screen_size),
+        method=tf.image.ResizeMethod.AREA)
+    return transformed_image.numpy().astype(np.uint8)
```
sguada: This would only work in Eager mode, in general we want the core of the library to work both in TF 1.x Graph and TF 2.x Eager.

sguada (changes requested):
Can we separate that change from the rest? I think the change to gym.core.Wrapper is good.

tf_agents/environments/atari_preprocessing.py
ageron: Sure, I'll send a separate PR. The motivation for this is that cv2 is only used here, and only to resize images, and since TF's resize function does the exact same thing, it feels like an unnecessary dependency (the one benefit is that cv2's implementation seems to be faster, at least on my machine).

tf_agents/environments/atari_preprocessing.py
ageron: I see, but that's not a regression, since the code was using cv2, which would not work in graph mode, right?

ageron: I reverted the cv2 change, I'll push a separate PR for that.

tf_agents/environments/atari_preprocessing.py
ageron: I reverted the cv2 change, I'll post a separate PR

seungjaeryanlee:
May I ask what the status of this PR is?

PR #120, opened 2019-05-24
Add a new Atari wrapper to fire on reset

ageron:
Inspired from [this code](https://github.com/openai/gym/blob/master/gym/wrappers/atari_preprocessing.py#L104).
Some Atari games do not start until you press fire.

### DIFF
--- a/tf_agents/environments/atari_wrappers.py
+++ b/tf_agents/environments/atari_wrappers.py
@@ -94,3 +94,19 @@ class AtariTimeLimit(wrappers.PyEnvironmentBaseWrapper):
   @property
   def game_over(self):
     return self._num_steps >= self._duration or self.gym.game_over
+
+
+class FireOnReset(gym.Wrapper):
+  """Start every episode with action 1 (FIRE) + another action (2).
+
+  This is required by some environments (e.g., Breakout) to actually start the
+  game.
+  """
+
+  def reset(self):
+    observation = self.env.reset()
+    action_meanings = self.env.unwrapped.get_action_meanings()
+    if action_meanings[1] == 'FIRE' and len(action_meanings) >= 3:
+        self.env.step(1)
+        observation, _, _, _ = self.env.step(2)
+    return observation
--- a/tf_agents/environments/atari_wrappers.py
+++ b/tf_agents/environments/atari_wrappers.py
@@ -99,12 +99,14 @@ class AtariTimeLimit(wrappers.PyEnvironmentBaseWrapper):
 class FireOnReset(gym.Wrapper):
   """Start every episode with action 1 (FIRE) + another action (2).
 
-  This is required by some environments (e.g., Breakout) to actually start the
-  game.
+  In some environments (e.g., BeamRider, Breakout, Tennis) nothing
+  happens until the player presses the FIRE button.
   """
 
   def reset(self):
     observation = self.env.reset()
+    # The following code is from https://github.com/openai/gym/...
+    # ...blob/master/gym/wrappers/atari_preprocessing.py
     action_meanings = self.env.unwrapped.get_action_meanings()
     if action_meanings[1] == 'FIRE' and len(action_meanings) >= 3:
         self.env.step(1)
--- a/tf_agents/environments/suite_atari.py
+++ b/tf_agents/environments/suite_atari.py
@@ -29,13 +29,14 @@ import gin.tf
 
 
 # Typical Atari 2600 Gym environment with some basic preprocessing.
-DEFAULT_ATARI_GYM_WRAPPERS = (atari_preprocessing.AtariPreprocessing,)
+DEFAULT_ATARI_GYM_WRAPPERS = (atari_preprocessing.AtariPreprocessing,
+                              atari_wrappers.FireOnReset)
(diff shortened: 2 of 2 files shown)


tf_agents/environments/atari_wrappers.py
```diff
@@ -94,3 +94,19 @@ def _step(self, action):
...
+  """Start every episode with action 1 (FIRE) + another action (2).
+
+  This is required by some environments (e.g., Breakout) to actually start the
+  game.
+  """
+
+  def reset(self):
+    observation = self.env.reset()
+    action_meanings = self.env.unwrapped.get_action_meanings()
+    if action_meanings[1] == 'FIRE' and len(action_meanings) >= 3:
```
egonina, line 111: can you add a comment about the hard-coded integesr (len >= 3) used here and env.step arguments?

ageron: I'd love to, but I'm really not sure. I just reused the same logic as in [this code](https://github.com/openai/gym/blob/master/gym/wrappers/atari_preprocessing.py#L104), which does not explain why it does that. I assume that some games require the fire button followed by any other action except NOOP (which is action 0). I'm not sure which ones, though. Would you like me to write that?

ageron:
I just ran a script that went through all Atari games provided by OpenAI gym, and it tried various things.

First, it tried doing nothing for 10,000 steps (action 0=NOOP), and checked whether anything at all changed (the observations, the rewards, the number of lives, game over, etc.). Just 3 games (out of 62) didn't change at all for 10,000 steps, not even by one pixel: BowlingNoFrameskip-v4, BreakoutNoFrameskip-v4 and AdventureNoFrameskip-v4. The first two get started when you FIRE (action 1). However, they freeze again after a life lost, so using FireOnReset only makes sense if you also terminate the episode when a life is lost.

Moreover, AdventureNoFrameskip-v4 only starts doing something with the sequence FIRE, ANY, OTHER where FIRE=1, ANY=any action whatsoever, OTHER=any action except FIRE or NOOP. This would explain the logic in the code, I suppose.

However, I tried doing FIRE, FIRE, action 2 (generally UP) on all games, just after `reset()`, followed by 10,000 NOOPs, and 15 games didn't change at all (but this time I didn't check for the observations, only rewards, game over or any life lost) during that time:

AdventureNoFrameskip-v4
AirRaidNoFrameskip-v4
BowlingNoFrameskip-v4
ChopperCommandNoFrameskip-v4
ElevatorActionNoFrameskip-v4
EnduroNoFrameskip-v4
HeroNoFrameskip-v4
MontezumaRevengeNoFrameskip-v4
NameThisGameNoFrameskip-v4
PitfallNoFrameskip-v4
RiverraidNoFrameskip-v4
SeaquestNoFrameskip-v4
SolarisNoFrameskip-v4
TutankhamNoFrameskip-v4
VideoPinballNoFrameskip-v4

Without this FireOnReset logic, only 3 more games do nothing:

BeamRiderNoFrameskip-v4
BreakoutNoFrameskip-v4
TennisNoFrameskip-v4

In other words, the FireOnReset wrapper is only useful for these 3 games (and only if you also terminate the episode on life lost).

So perhaps this wrapper is not that useful after all. It adds some extra logic while only being useful in very rare cases.

Wdyt?

tf_agents/environments/atari_wrappers.py
egonina: ok, no problem. you can leave out the comment since it comes from a the gym code

egonina:
I'd say leave this out since it's only partially useful in a few games. But I'll let others comment who are more familiar with Atari CC @sguada @oars @samfishman

tf_agents/environments/atari_wrappers.py
samfishman: Add a comment referencing the code you lifted this from?

tf_agents/environments/atari_wrappers.py
```diff
@@ -94,3 +94,19 @@ def _step(self, action):
   @property
   def game_over(self):
     return self._num_steps >= self._duration or self.gym.game_over
+
+
+class FireOnReset(gym.Wrapper):
+  """Start every episode with action 1 (FIRE) + another action (2).
+
+  This is required by some environments (e.g., Breakout) to actually start the
```
samfishman: It's not actually required - a DQN agent can solve breakout perfectly well without this wrapper! Especially if you evaluate with an epsilon-greedy policy (with small epsilon). So I'd reword this to mention that it can be helpful in the 3 games you mention, but that it is not completely necessary.

tf_agents/environments/suite_atari.py
```diff
@@ -29,13 +29,14 @@
 
 
 # Typical Atari 2600 Gym environment with some basic preprocessing.
-DEFAULT_ATARI_GYM_WRAPPERS = (atari_preprocessing.AtariPreprocessing,)
+DEFAULT_ATARI_GYM_WRAPPERS = (atari_preprocessing.AtariPreprocessing,
+                              atari_wrappers.FireOnReset)
```
samfishman: Remove this from the defaults.

PR #124, opened 2019-05-27
Make the use of tf_functions optional

julianstastny:
I had a bug where using a slightly customized version of the code from `agents/ppo/examples/v2/train_eval.py` resulted in extreme slow-down whenever the script entered `metric_utils.eager_compute` in line 205.

Since then I identified the use of `common.function(driver.run)` in line 156 of `eval/metric_utils.py` as the reason for this. Removing common.function fixed the issue, which is why I would suggest to make the usage optional until the cause of slow-down is identified and fixed.

### DIFF
--- a/tf_agents/eval/metric_utils.py
+++ b/tf_agents/eval/metric_utils.py
@@ -123,7 +123,8 @@ def eager_compute(metrics,
                   num_episodes=1,
                   train_step=None,
                   summary_writer=None,
-                  summary_prefix=''):
+                  summary_prefix='',
+                  use_tf_functions=True):
   """Compute metrics using `policy` on the `environment`.
 
   *NOTE*: Because placeholders are not compatible with Eager mode we can not use
@@ -153,7 +154,11 @@ def eager_compute(metrics,
       policy,
       observers=metrics,
       num_episodes=num_episodes)
-  common.function(driver.run)(time_step, policy_state)
+  
+  if use_tf_functions:
+    common.function(driver.run)(time_step, policy_state)
+  else:
+    driver.run(time_step, policy_state)
 
   results = [(metric.name, metric.result()) for metric in metrics]
   # TODO(b/120301678) remove the summaries and merge with compute


egonina:
hm, so using tf_function adds autodeps, for the code in the driver that is written eagerly to execute the correct dependencies, otherwise the code won't do the right thing as some things will be skipped that are not on the direct dependency path of the final output tensor. Did you verify that the results were correct if you don't use tf_function? I think the right thing is to figure out why we get the slow down rather than not use tf_function

julianstastny:
Thanks for making me aware of this, I didn't know that, as I used PyTorch up until now.
Looking at the average return over time (also in the eval, where this issue is relevant) it seems to work even if not using tf_function. However, I am still struggling to get my Gym-Wrapper for Visualization to be compatible with tf-agents, so I can't say for sure.

Generally I agree that it it is the right thing to solve the underlying reason for slowdown, but right now the slowdown makes it unusable (been waiting for over an hour and nothing seems to happen), which is why I suggested the patch. If no one else experiences the problem (at least I saw no issue where there was such extreme slowdown), then there is of course no need for that.

julianstastny:
So, I found out that 
a) the results are correct despite not using tf-functions in metric_utils.py
b) Weirdly, using tf-functions in train_eval.py works (and makes the script significantly faster), while using it in metric_utils.py still leads to freezing.
Maybe this gives you some useful information for solving the slow-down issue.

egonina:
Are you hitting this tf_function call in eager_compute call here: https://github.com/tensorflow/agents/blob/f2276c6a36b68a3a8d510d846a4cc0338a5ae1f1/tf_agents/eval/metric_utils.py#L156 ?

julianstastny:
Yes, and if I do, it freezes the program.
Changing the line to `driver.run(time_step, policy_state)` fixed it.

tfboyd:
Last activity was on June 5th.  Closing due to the amount of time that has passed and probability of this finding a path to being merged.  This clean up decision can be revisited if desired.  Just ping me in the comments.

PR #131, opened 2019-06-06
Replace cv2.resize() with tf.image.resize()

ageron:
This removes the OpenCV dependency. Since cv2 is a bit faster than TensorFlow for this method (on my machine, at least), we may want to use cv2 if it's available, and fallback to TF if not. Wdyt?

### DIFF
--- a/tf_agents/environments/atari_preprocessing.py
+++ b/tf_agents/environments/atari_preprocessing.py
@@ -33,7 +33,7 @@ from __future__ import print_function
 import gin
 from gym.spaces.box import Box
 import numpy as np
-import cv2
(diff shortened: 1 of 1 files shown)


tfboyd:
Closing due to lack of activity.  I am some what out of my element (cannot state this strongly enough), but from past experience being in the room changing the input pipeline is always a little scary. With a variety of subtle options when it comes to doing a resize there is always a concern the results between two seemingly identical calls will actually be the same and if the resulting difference will impact training beyond a possible throughput gain or loss.  Sorry this was not looked at and thank you for the idea.

PR #138, opened 2019-06-13
fix a bug when frame_skip==1

hnyu:
When self.frame_skip==1, the original code will fill in buffer[1]. However, inside self._pool_and_resize(), buffer[0] is always used. So the entire episode will always output the first frame as the observations at different time steps. I add a special handling for the case of self.frame_skip==1 to fix this issue.

### DIFF
--- a/tf_agents/environments/atari_preprocessing.py
+++ b/tf_agents/environments/atari_preprocessing.py
@@ -187,7 +187,11 @@ class AtariPreprocessing(object):
       # We max-pool over the last two frames, in grayscale.
       elif time_step >= self.frame_skip - 2:
         t = time_step - (self.frame_skip - 2)
+        # when frame_skip==1, self.screen_buffer[1] will be filled!
         self._fetch_grayscale_observation(self.screen_buffer[t])
+
+    if self.frame_skip == 1:
+      self.screen_buffer[0] = self.screen_buffer[1]
 
     # Pool the last two observations.
     observation = self._pool_and_resize()


hnyu:
I signed it!

tfboyd:
@sguada @ebrevdo @kbanoop   Does this fix a bug?  The change is small enough and there is no conflict that I would like a quick review before I close due to inactivity.

tf_agents/environments/atari_preprocessing.py
```diff
@@ -187,8 +187,12 @@ def step(self, action):
       # We max-pool over the last two frames, in grayscale.
       elif time_step >= self.frame_skip - 2:
         t = time_step - (self.frame_skip - 2)
```
samfishman, line 189: Instead of special-casing `frame_skip == 1` below, let's redefine `t` as `t = time_step - max(self.frame_skip - 2, 0)` and add a comment that this ensures proper behavior for `frame_skip == 1`.

samfishman (changes requested):
Thanks for finding this bug @hnyu! And apologies for letting this fall through the cracks. Thanks to @tfboyd for bringing this back to our attention :).

tf_agents/environments/atari_preprocessing.py
hnyu: Good suggestion! Sorry that somehow this fork is gone on my side. Is there any other way that I can update this PR?

samfishman: Hmm, I don't think there's a way to update this PR, but if you create a new one, I can close this one and link to the new one.

hnyu: @samfishman I've created a new PR #302 . Please close this one, thanks!

tfboyd:
See new PR #302 .

PR #142, opened 2019-06-18
ValueError: Not a same graph

seungjaeryanlee:
I successfully converted observation spec from `BoundedArraySpec` to `BoundedTensorSpec`! However, I get a ValueError when I try to run `train_eval_rnd.py`:

```
ValueError: Tensor("RNDNetwork/EncodingNetwork/EncodingNetwork/dense/kernel/Read/ReadVariableOp:0", 
shape=(4, 100), dtype=float32) must be from the same graph as 
Tensor("RNDNetwork/EncodingNetwork/flatten/Reshape:0", shape=(1, 4), dtype=float32).
```

Do you have any suggestions on where I should pay attention to? I have been trying a few things but none of them fixed the issue.

Here is the full output: [output.txt](https://github.com/tensorflow/agents/files/3299269/output.txt)


seungjaeryanlee:
Wrong destination branch: sorry for the noise

PR #151, opened 2019-07-02
Fix the output shape of ValueRnnNetwork

### LINKED ISSUE #87: ValueRnnNetwork added extra dimention when input does not have time dimension
In tf_agents.networks.value_rnn_network.ValueRnnNetwork.call(), if `observation` does not have time dimension, it is expanded one more dimension for time at
https://github.com/tensorflow/agents/blob/82317735484b99f8bcb6ec7cb99d1e73f058ca55/tf_agents/networks/value_rnn_network.py#L153-L158

And the `value` just before the following line is a flattened version. Then it's unflattened by the following line to a shape with an extra time dimension:
https://github.com/tensorflow/agents/blob/82317735484b99f8bcb6ec7cb99d1e73f058ca55/tf_agents/networks/value_rnn_network.py#L184

Shouldn't the correct behavior be that if the input observation does not have time dimension, the output should not either?

emailweixu:
Fix #87

### DIFF
--- a/tf_agents/networks/value_rnn_network.py
+++ b/tf_agents/networks/value_rnn_network.py
@@ -181,5 +181,6 @@ class ValueRnnNetwork(network.Network):
 
     value = self._value_projection_layer(states)
     value = tf.reshape(value, [-1])
-    value = batch_squash.unflatten(value)
+    if has_time_dim:
+      value = batch_squash.unflatten(value)
     return value, network_state


tfboyd:
There has been a lot of refactoring.  While I did not test it, looking at the code I believe the issue is fixed by this line:
https://github.com/tensorflow/agents/blob/master/tf_agents/networks/lstm_encoding_network.py#L205

This PR was also from a long time ago.  We are trying to do a much better job of handling PRs and int he past two weeks have been much more responsive.  We hope to keep it up and sorry we let this hang, this was 100% a fix; and thank you for pointing it out.

PR #172, opened 2019-07-24
Remove duplicate loss functions defined in dqn_agent

seungjaeryanlee:
The two function `element_wise_squared_loss` and `element_wise_huber_loss` is already in `utils/commons.py`, so redefining them in `dqn_agent.py` seems unnecessary.

https://github.com/tensorflow/agents/blob/96ca00544ca1f6524c89194afe272ebf07773b05/tf_agents/utils/common.py#L966

https://github.com/tensorflow/agents/blob/96ca00544ca1f6524c89194afe272ebf07773b05/tf_agents/utils/common.py#L971

### DIFF
--- a/tf_agents/agents/dqn/dqn_agent.py
+++ b/tf_agents/agents/dqn/dqn_agent.py
@@ -66,18 +66,6 @@ class DqnLossInfo(collections.namedtuple('DqnLossInfo',
   pass
 
 
-# TODO(damienv): Definition of those element wise losses should not belong to
-# this file. Move them to utils/common or utils/losses.
-def element_wise_squared_loss(x, y):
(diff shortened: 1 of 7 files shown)


kbanoop:
Can you please change the examples as well, e.g. https://github.com/tensorflow/agents/blob/b08a142edf180325b63441ec1b71119c393c4a64/tf_agents/agents/dqn/examples/v2/train_eval.py#L172

seungjaeryanlee:
Thank you for reviewing @kbanoop !

I replaced all places where dqn_agent loss functions are used. There seems to be some documentation files left. Are these auto-generated?

kbanoop:
Yes the docs are programmatically generated, but we have to run the scripts to do that manually. Thanks for the cleanup!

PR #173, opened 2019-07-25
fix code style

pana1990:
camel case to snake case

### DIFF
--- a/tf_agents/metrics/tf_metrics_test.py
+++ b/tf_agents/metrics/tf_metrics_test.py
@@ -29,7 +29,7 @@ from tensorflow.python.eager import context  # TF internal
 
 class TFDequeTest(tf.test.TestCase):
 
-  def testDataIsZero(self):
+  def test_data_is_zero(self):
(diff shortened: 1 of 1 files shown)


pana1990:
I signed it!

kbanoop (commented):
Thanks for the contribution! I agree we usually follow snake_case for function names, but for some reason, we use CamelCase for function names in our tests through most of the codebase. So I would rather keep it as it is for consistency

pana1990:
@kbanoop thanks for your reply, In my honest opinion we should follow a consistent standard for the code style like STYLE_GUIDE.md say "TF Agents Style Guide say "Follow the TensorFlow style guide"

sorry for my poor english in advanced :/

kbanoop:
Agree, but TensorFlow also follows the same `CamelCase for test names` style. 

See e.g. https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/data/kernel_tests/flat_map_test.py

PR #196, opened 2019-08-28
Tutorial Fixes

jackdreilly:
1. `time_step` needs to be reset after each episode.
2. `next_time_step` should be `time_step` for collecting rewards.


jackdreilly:
@googlebot I signed it!

Issue #198, opened 2019-08-29
Using the add_batch method from PyUniformReplayBuffer throws an IndexError: tuple index out of range.


j0rd1smit:
Using the `add_batch` method from `PyUniformReplayBuffer` throws an `IndexError: tuple index out of range`. When you do something simliar with an `TFEnv`  all the trajectories are batched but this is not the case in `PyEnv`. I think that is why the observer `lambda x: buffer.add_batch(batch_nested_array(x))` works and the observer  `buffer.add_batch` doesn't. Bellow are some examples codes.

This doesn't work:

    env = suite_gym.load('CartPole-v0')  
    policy = random_py_policy.RandomPyPolicy(time_step_spec=env.time_step_spec(),action_spec=env.action_spec())  
    buffer = PyUniformReplayBuffer(policy.trajectory_spec, 1000)  
      
    observers = [buffer.add_batch]  
    driver = py_driver.PyDriver(env, policy, observers, max_steps=1, max_episodes=1)  
      
    initial_time_step = env.reset()  
    final_time_step, _ = driver.run(initial_time_step)

This works:

    env = suite_gym.load('CartPole-v0')  
    policy = random_py_policy.RandomPyPolicy(time_step_spec=env.time_step_spec(),action_spec=env.action_spec())  
    buffer = PyUniformReplayBuffer(policy.trajectory_spec, 1000)  
      
    observers = [lambda x: buffer.add_batch(batch_nested_array(x))] #Batched the traj
    driver = py_driver.PyDriver(env, policy, observers, max_steps=1, max_episodes=1)  
      
    initial_time_step = env.reset()  
    final_time_step, _ = driver.run(initial_time_step)

I think the best way to solve this is batching the trajactory in `PyDriver` to stay consistent with the `TfDriver`. This would only result into a single change in the `.run()` method from `PyDriver`.
original:

    traj = trajectory.from_transition(time_step, action_step, next_time_step)
Proposed replacement:

    traj = trajectory.from_transition(time_step, action_step, next_time_step)  
	traj = nest_utils.batch_nested_array(traj)

sguada:
To use add_batch the data should be batched, i.e. the environment should be batched. This should work:
```
env = batched_py_environment.BatchedPyEnvironment([suite_gym.load('CartPole-v0')])
policy = random_py_policy.RandomPyPolicy(time_step_spec=env.time_step_spec(),action_spec=env.action_spec())  
[...]

PR #200, opened 2019-08-29
Fixed: add_batch method  observer on PyUniformReplayBuffer bug #198

j0rd1smit:
Fixes issue #198

### DIFF
--- a/tf_agents/drivers/py_driver.py
+++ b/tf_agents/drivers/py_driver.py
@@ -22,6 +22,7 @@ from __future__ import print_function
 import numpy as np
 from tf_agents.drivers import driver
 from tf_agents.trajectories import trajectory
+from tf_agents.utils import nest_utils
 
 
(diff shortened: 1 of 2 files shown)


nealwu:
Hi, this change breaks a test on https://github.com/tensorflow/agents/blob/master/tf_agents/agents/ppo/examples/v1/train_eval.py, specifically because of metric_utils.py, which uses py_driver: https://github.com/tensorflow/agents/blob/master/tf_agents/eval/metric_utils.py#L72.

The error message is
```
ValueError: Batch size 1 does not match previously set batch size 2. Make sure your batch size is set correctly in BatchedPyMetric initialization and that the batch size remains constant.
```

@ebrevdo, do you know what the right fix would be?

tfboyd:
Closing due to time passed.  The mismatch between py and tf version of components is confusing; and is being improved slowly based on customer need.

PR #204, opened 2019-09-07
Fix ActionRepeat to preserve the reward dtype and to not repeat actions on an implicit reset.

alexlee-gk:
E.g. float32 rewards were being converted to float64 before the fix.


oars:
Great catches! Can you also update the tests.

oars:
Pushed in: 1dd93823c87289f9e5914fb13ea2eb8bc844be5a Thanks @alexlee-gk

PR #224, opened 2019-10-18
close envs in ParallelPyEnvironment when it closes

witwolf:
https://github.com/HorizonRobotics/alf/issues/206

In some scenarios, env itself is a  process or  multi-process,  ParallelPyEnvironment.close will hang , 
and resource will not be released


witwolf:
@googlebot I signed it

PR #243, opened 2019-11-11
Add trajectory arg to call fn for CounterMetric

krishpop:
Added trajectory as argument for CounterMetric.call to enable compatibility with TFPyMetric wrapper.

### DIFF
--- a/tf_agents/metrics/py_metrics.py
+++ b/tf_agents/metrics/py_metrics.py
@@ -261,7 +261,8 @@ class CounterMetric(py_metric.PyMetric):
   def reset(self):
     self._np_state.count = np.int64(0)
 
-  def call(self):
+  def call(self, trajectory):
+    del trajectory  # unused
     self._np_state.count += 1
 
   def result(self):


kbanoop:
I think this is intended behavior. PyMetrics are allowed to take any or no argument, whereas PyStepMetrics take trajectories:
https://github.com/tensorflow/agents/blob/08a30a5f7a434dbd3646d4cd8402258a665a44cc/tf_agents/metrics/py_metric.py#L179
And they are also used in a few places such as:
https://github.com/tensorflow/agents/blob/08a30a5f7a434dbd3646d4cd8402258a665a44cc/tf_agents/agents/dqn/examples/v1/train_eval_atari.py#L350 

But I agree we should make it compatible with TFPyMetric. One way is to generalize the TFPyMetric to support other arguments besides trajectory. Otherwise, it may be pretty easy to implement the Counter metric in TF directly.

ebrevdo:
@kbanoop do you recommend we close this pull request?  @krishpop is Anoop's answer sufficient to revert this?

kbanoop:
I think we can close for now. @krishpop  please feel free to reopen if you have more questions.

PR #252, opened 2019-11-18
Eps Greedy Policy fix for policy info containing tensors of rank > 1.

mhe500:
Semantics of tf.where in v2 differ from tf.compat.v1.where, as condition
rank must match true/false tensors. Addressed by adding
nest_utils.where2 rather than reverting to tf.compat.v1 so that
condition can eventually have rank > 1 (as suggested by
TODO(b/133175894).

### DIFF
--- a/tf_agents/utils/nest_utils.py
+++ b/tf_agents/utils/nest_utils.py
@@ -556,3 +556,26 @@ def where(condition, true_outputs, false_outputs):
   """
   return tf.nest.map_structure(lambda t, f: tf.compat.v1.where(condition, t, f),
                                true_outputs, false_outputs)
+
+def wherev2(condition, true_outputs, false_outputs):
+  """Generalization of tf.where where supporting nests as the outputs.
+
+
+  Args:
+    condition: A boolean Tensor of shape [B,].
+    true_outputs: Tensor or nested tuple of Tensors of any dtype, each with
+      shape [B, ...], to be split based on `condition`.
+    false_outputs: Tensor or nested tuple of Tensors of any dtype, each with
+      shape [B, ...], to be split based on `condition`.
+
+  Returns:
+    Interleaved output from `true_outputs` and `false_outputs` based on
+    `condition`.
+  """
+  return tf.nest.map_structure(
+    lambda t, f: tf.where(
+      # Under semantics of tf.where v2, cond must have same rank as a/b. Differs from tf.compat.v1.where.
+      tf.reshape(condition,
+                 tf.concat([tf.shape(condition), tf.ones(tf.rank(t) - tf.rank(condition), dtype=tf.int32)], axis=0)),
+      t, f),
+    true_outputs, false_outputs)
--- a/tf_agents/utils/nest_utils.py
+++ b/tf_agents/utils/nest_utils.py
@@ -560,10 +560,10 @@ def where(condition, true_outputs, false_outputs):
     Interleaved output from `true_outputs` and `false_outputs` based on
     `condition`.
   """
-  def where_with_matching_ranks(t, f):
+  def _where_with_matching_ranks(t, f):
       rank_difference = tf.rank(t) - tf.rank(condition)
       condition_shape = tf.concat([tf.shape(condition), tf.ones(rank_difference, dtype=tf.int32)], axis=0)
       reshaped_condition = tf.reshape(condition, condition_shape)
       return tf.where(reshaped_condition, t, f)
 
-  return tf.nest.map_structure(where_with_matching_ranks, true_outputs, false_outputs)
+  return tf.nest.map_structure(_where_with_matching_ranks, true_outputs, false_outputs)
--- a/tf_agents/utils/nest_utils.py
+++ b/tf_agents/utils/nest_utils.py
@@ -560,9 +560,10 @@ def where(condition, true_outputs, false_outputs):
     Interleaved output from `true_outputs` and `false_outputs` based on
     `condition`.
   """
-  return tf.nest.map_structure(
-    lambda t, f: tf.where(
-      tf.reshape(condition,
-                 tf.concat([tf.shape(condition), tf.ones(tf.rank(t) - tf.rank(condition), dtype=tf.int32)], axis=0)),
-      t, f),
-    true_outputs, false_outputs)
+  def where_with_matching_ranks(t, f):
+      rank_difference = tf.rank(t) - tf.rank(condition)
+      condition_shape = tf.concat([tf.shape(condition), tf.ones(rank_difference, dtype=tf.int32)], axis=0)
+      reshaped_condition = tf.reshape(condition, condition_shape)
+      return tf.where(reshaped_condition, t, f)
+
+  return tf.nest.map_structure(where_with_matching_ranks, true_outputs, false_outputs)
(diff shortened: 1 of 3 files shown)


tf_agents/utils/nest_utils.py
```diff
@@ -556,3 +556,26 @@ def where(condition, true_outputs, false_outputs):
   """
   return tf.nest.map_structure(lambda t, f: tf.compat.v1.where(condition, t, f),
                                true_outputs, false_outputs)
+
+def wherev2(condition, true_outputs, false_outputs):
```
egonina: can you rename to where_nest or something similar since it's confusing what v2 means in this case?

tf_agents/utils/nest_utils.py
```diff
@@ -556,3 +556,26 @@ def where(condition, true_outputs, false_outputs):
...
+  """Generalization of tf.where where supporting nests as the outputs.
```
egonina: and inputs?

tf_agents/utils/nest_utils_test.py
```diff
@@ -680,6 +680,20 @@ def testWhere(self):
     expected = (np.array([0, 1, 1, 0, 1]), np.array([1, 7, 8, 4, 10]))
     self.assertAllEqual(expected, result)
 
+  def testWhereV2(self):
```
egonina: can you test with empty tensors to ensure the right error message etc?

egonina (commented):
thanks for the PR! added some comments

tf_agents/utils/nest_utils.py
mhe500: Hi @egonina . Unfortunately nest_utils.where() already exists and uses tf.compat.v1.where. 

One option could be to completely **replace** that method with the new one. Given the new method expands the condition's rank with size 1 dimensions to match t/f, I believe it **should** be backward compatible. I chose not to do this because, in the end the behavior **is** different, and this is a "public" function. That said, given that TFA is undergoing rapid change, on second thought it seems like an acceptable risk.

@egonina, what would be your thoughts on completely replacing the existing nest_utils.where() with this version of the method?

tf_agents/utils/nest_utils.py
mhe500: Yes, in that the inputs are the nests from which tf.where selects elements to construct the output. I figured this is sort of implicit in the nature of "where" (and, honestly, I copied the docstring from the existing nest_utils.where() :) ). 

I will update the docstring to be clearer.

tf_agents/utils/nest_utils_test.py
mhe500: Sure. Let me think about what tests make the most sense here and I will update.

tf_agents/utils/nest_utils.py
egonina: what do you mean the behavior is different? we could have one method that handles both types of input? I would prefer that than having two separate ones.

mhe500: Hi @egonina.

Let me clarify. By "behavior is different" I meant that under TensorFlow 2.0, the behavior `tf.where()` is _not_ exactly the same as `tf.compat.v1.where()`. For example, under the TensorFlow 1.0 version, the condition must be a **vector**, but under TensorFlow 2.0 the condition may be a **multi-dimensional Tensor** (this is a primary difference).

The existing TF-Agents `nest_utils.where()` uses `tf.compat.v1.where()` (the old behavior). If we completely replace `nest_utils.where()` with my new version, under the hood it will use TensorFlow 2.0's `tf.where()`.

While I believe that the way I have written this, it is backward compatible, the overall behavior is, nonetheless _different_. As one example, if you passed a `condition` argument that is a multi-dimensional Tensor to the existing `nest_utils.where()` you would get a `ValueError` exception because `tf.compat.v1.where()` does not allow this. However, in my version, this would be acceptable because `tf.where()` (under Tensorflow 2.0) allows this.

All that said, even though there is a difference, I think the overall risk of completely replacing `nest_utils.where()` with the proposed version is very small and will allow enhancements in the future which leverage the new capabilities of `tf.where()` under TensorFlow 2.0.

Please let me know if that clarifies things.
mhe

egonina: Thanks for the clarification! I am fine replacing the util to use TF 2.0 where, as long as our tests pass, I think we should be fine. We are moving away from 1.0 behavior. 

@ebrevdo @sguada what do you think?

mhe500:
@googlebot I fixed it.

tf_agents/utils/nest_utils.py
mhe500: Hi @egonina. I've updated the code to a have a single version of `nest_utils.where` using TensorFlow 2.0's `tf.where`.

tf_agents/utils/nest_utils.py
mhe500: Clarified.

tf_agents/utils/nest_utils_test.py
mhe500: Added tests for different ranks and same rank different dimension tensore. Existing test for `nest_utils.where` only covered same rank/same dimension.

tf_agents/utils/nest_utils.py
```diff
@@ -554,5 +560,9 @@ def where(condition, true_outputs, false_outputs):
     Interleaved output from `true_outputs` and `false_outputs` based on
     `condition`.
   """
-  return tf.nest.map_structure(lambda t, f: tf.compat.v1.where(condition, t, f),
-                               true_outputs, false_outputs)
+  return tf.nest.map_structure(
+    lambda t, f: tf.where(
+      tf.reshape(condition,
+                 tf.concat([tf.shape(condition), tf.ones(tf.rank(t) - tf.rank(condition), dtype=tf.int32)], axis=0)),
```
egonina: i wonder if there's a cleaner way to rewrite this part here so its not all in one line? maybe define a helper _new_shape() and pass that to reshape?

tf_agents/utils/nest_utils_test.py
egonina: great, thanks!

tf_agents/utils/nest_utils.py
egonina: great!

tf_agents/utils/nest_utils.py
mhe500: Hi @egonina . I've converted the lambda into a function and broken it into multiple lines for clarity per your suggestion.

tf_agents/utils/nest_utils.py
```diff
@@ -554,5 +560,10 @@ def where(condition, true_outputs, false_outputs):
     Interleaved output from `true_outputs` and `false_outputs` based on
     `condition`.
   """
-  return tf.nest.map_structure(lambda t, f: tf.compat.v1.where(condition, t, f),
-                               true_outputs, false_outputs)
+  def _where_with_matching_ranks(t, f):
```
egonina: thank you for refactoring this part of the code. can you lastly add a comment about what this helper function does?

egonina:
Thanks for addressing the comments and the PR! I added another small comment about documentation, otherwise LGTM and we can merge

tf_agents/utils/nest_utils.py
mhe500: Hi @egonina , see the docstring I added to the helper. Thanks for all your feedback.

oars:
Thank you for the contribution! I cleaned it up a bit to make all tests pass, formatting, etc.

PR #269, opened 2019-12-11
Support for 'training' flag (for dropout) in DQN Agent.

mhe500:
Pass 'training' boolean to q_network/q_rnn_network/categorical_q_network and propagate as appropriate. Also capture q_network.losses into DQNAgent's loss before the q_network is called again to calculate Ddqn targets and the losses are reset.


ebrevdo:
I'm adding this to all agents, after which perhaps you could rebase and add in your unit tests?

mhe500:
Sure, I'd be happy to.

PR #272, opened 2019-12-16
added editorconfig to help to follow the tensorflow style

pana1990:


tfboyd:
I like this idea and appreciate the PR and time it takes to create one.  Unfortunately, I do not think we want to include this in the repository.  I would be more inclined if the based Tensorflow project had the file.

Closing PRs and not merging them is the least fun thing I have to do.

pana1990:
@tfboyd  ok i will trying send the same pr to tensorflow, and if they like this enh i come back to create pr again, right?

tfboyd:
If you are successful, **I will 100% try to talk the team into it** using the tensorflow project as the precedent. I do not like saying no.  I also checked some other projects, .e.g pytorch, and they also do not have an editorconfig.  I was looking for some data points.  I want you to know I did some looking around before I closed your contribution.  

Thank you for the nice comment.

PR #278, opened 2019-12-29
Remove ValueError of q_value dimension in ddpg_agent critic_loss

cjy1992:
When using a critic_rnn_network, the output q_value will have shape [batch_size, time,]. However, the current ddpg_agent will raise ValueError in critic_loss for q_value with time dimension. Remove this ValueError part should solve the problem.


cjy1992:
@googlebot I signed it!

tfboyd:
@kbanoop  Can you take a quick look and see if this is a problem that needs fixed and if this fixes the problem in an acceptable way?

@cjy1992  Sorry for the delay.  We are starting a new process of dealing with PRs faster and transparently.  Sometimes we get great changes that are more difficult to accept than it seems.  Rather than let them sit, we are going to try and be transparent and indicate we cannot merge them.  On the plus side, things we can merge I hope we get those done faster.  Thank you for the contribution, we very much appreciate the time it takes to contribute and appreciate your interest.  I hope we get this to a conclusion in the next few days.

tfboyd:
We saved a few back in Feb and we have done a lot better since then, but this one just didn't make it. I am sorry. Closing to accept reality.

Issue #276, opened 2019-12-25
Epsilon greedy policy error when generating new action from Dict Action space. 


JaCoderX:
I'm trying to convert a custom gym project (called BTgym) to work as a tf-agent env.
the original observation space and the action space are both `gym.spaces.Dict`.
but for the moment I have simplified the observation space so I can fit the env to run using the same code of DQN tutorial example (as a proof of concept). so the modified spaces are as follows:
```
Observation Spec:
BoundedTensorSpec(shape=(6, 1, 5), dtype=tf.float32, name='observation/external', minimum=array(-100., dtype=float32), maximum=array(100., dtype=float32))
Action Spec:
OrderedDict([('default_asset', BoundedTensorSpec(shape=(), dtype=tf.int64, name='action/default_asset', minimum=array(0), maximum=array(3)))]) 
```

Error occur under **Training the agent** section, when performing `collect_step()`:
```
Traceback (most recent call last):
  File "/home/jack/envTest.py", line 251, in <module>
    collect_step(train_env, agent.collect_policy, replay_buffer)
  File "/home/jack/envTest.py", line 209, in collect_step
    action_step = policy.action(time_step)
  File "/home/jack/tf_agents/policies/tf_policy.py", line 278, in action
    step = action_fn(time_step=time_step, policy_state=policy_state, seed=seed)
  File "/home/jack/tf_agents/utils/common.py", line 131, in with_check_resource_vars
    return fn(*fn_args, **fn_kwargs)
  File "/home/jack/tf_agents/policies/epsilon_greedy_policy.py", line 106, in _action
    random_action.action)
  File "/home/jack/anaconda3/envs/deep/lib/python3.6/site-packages/tensorflow_core/python/util/dispatch.py", line 180, in wrapper
    return target(*args, **kwargs)
  File "/home/jack/anaconda3/envs/deep/lib/python3.6/site-packages/tensorflow_core/python/ops/array_ops.py", line 3753, in where
    return gen_math_ops.select(condition=condition, x=x, y=y, name=name)
  File "/home/jack/anaconda3/envs/deep/lib/python3.6/site-packages/tensorflow_core/python/ops/gen_math_ops.py", line 9430, in select
    condition, x, y, name=name, ctx=_ctx)
  File "/home/jack/anaconda3/envs/deep/lib/python3.6/site-packages/tensorflow_core/python/ops/gen_math_ops.py", line 9462, in select_eager_fallback
[...]

PR #301, opened 2020-02-06
added support for Dict action space

JaCoderX:
Epsilon greedy policy error when generating new action from Dict Action space. #276

### DIFF
--- a/tf_agents/agents/dqn/dqn_agent.py
+++ b/tf_agents/agents/dqn/dqn_agent.py
@@ -507,7 +507,6 @@ class DqnAgent(tf_agent.TFAgent):
 
   def _compute_q_values(self, time_steps, actions, training=False):
     network_observation = time_steps.observation
-    actions = tf.nest.flatten(actions)[0]
 
     if self._observation_and_action_constraint_splitter is not None:
       network_observation, _ = self._observation_and_action_constraint_splitter(
@@ -517,7 +516,7 @@ class DqnAgent(tf_agent.TFAgent):
                                   training=training)
     # Handle action_spec.shape=(), and shape=(1,) by using the multi_dim_actions
     # param. Note: assumes len(tf.nest.flatten(action_spec)) == 1.
-    multi_dim_actions = tf.nest.flatten(self._action_spec)[0].shape.rank > 0
+    multi_dim_actions = self._action_spec.shape.rank > 0
     return common.index_with_actions(
         q_values,
         tf.cast(actions, dtype=tf.int32),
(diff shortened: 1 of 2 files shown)


JaCoderX:
@googlebot I signed it!

tfboyd:
@oars I saw you did a +2 on the internal sync, so I added you as the reviewer here, you can assign someone else.

ebrevdo:
We're trying to tighten up the contract that Agents present to the world, in particular we don't want to allow arbitrary nested structures "so long as its only got one entry", if the agent does not support more than one action in the action space.  Better if you just add a wrapper that converts the tensor to a dict.

If we start allowing this kind of soft contract, it makes it much harder for future developers to know which agents *really* support multiple action spaces; and makes it harder for us to add pytype to tf-agents.  Sorry; will have to reject this PR.

ebrevdo:
(a good followup PR would be to disallow nested action_specs in this agent in the first place).

JaCoderX:
> We're trying to tighten up the contract that Agents present to the world, in particular we don't want to allow arbitrary nested structures "so long as its only got one entry", if the agent does not support more than one action in the action space.

@ebrevdo, I agree with you that the code need to be simple and clean for the agents, especially if a certain agent implementation (this case DQN) is meant for just one dimensional action space.

but this PR contain 2 commits, one for the DQN and the second is for the epsilon greedy policy. 
As far as I understand, the epsilon greedy policy isn't bounded only for DQN and can be used by other agents that can have a structured action space? is that commit also too case specific?

personally I just wanted to test the tf-agent framework with an external and complex gym project and see they work well together. I use the DQN tutorial just for an easy entry to tf-agent framework.

anyway thanks for helping out :)

ebrevdo:
Ah I didn't see the changes to epsilon_greedy.  That change is fine!  Can you revert the change to DQN agent?

tf_agents/agents/dqn/dqn_agent.py
```diff
@@ -546,7 +547,8 @@ def _compute_next_q_values(self, next_time_steps):
     # action constraints are respected and helps centralize the greedy logic.
     greedy_actions = self._target_greedy_policy.action(
         next_time_steps, dummy_state).action
-
+    greedy_actions = tf.nest.flatten(greedy_actions)[0]
```
ebrevdo: revert this file please.

JaCoderX:
DQN changes had been reverted

ebrevdo (approved):

ebrevdo:
Is there a unit test?

tfboyd:
@JacobHanouna  If there is a unit test, please reference it. If not can you add one and we can merge this can get it closed. Not to be a drag, but without a test the chances of it getting broken by some other change increases greatly.  Thank you for understanding.  We are **so close** on this and appreciate you powering through to the end.

JaCoderX:
[Tutorial: Integration with TF-Agents RL Framework](https://github.com/Kismuz/btgym/issues/133)

tfboyd:
So close to high fives all around.

tfboyd:
MERGED

PR #311, opened 2020-02-19
Make the NumEpisodesObserver count all episodes in a batch of trajectories  

leroidauphin:
The `NumStepsTransitionObserver` counts the number of steps in a batch of trajectories by using a `reduce_sum` across the batch dimension. By contrast, the `NumEpisodesObserver` only counts the number of episodes in the zero-th indexed member of the batch. 

This PR modifies the `NumEpisodesObserver` to follow the pattern established in the `NumStepsTransitionObserver` and use `reduce_sum` to count episodes across all of the episodes in the batch of trajectories.

### DIFF
--- a/tf_agents/drivers/test_utils_test.py
+++ b/tf_agents/drivers/test_utils_test.py
@@ -1,0 +1,58 @@
+# coding=utf-8
+# Copyright 2020 The TF-Agents Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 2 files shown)


leroidauphin:
@googlebot I signed it!

egonina:
Thanks for the PR! Would you mind adding a unit test to ensure this behaves as intended?

leroidauphin:
> Thanks for the PR! Would you mind adding a unit test to ensure this behaves as intended?

My apologies - I have now added a unit test.

tf_agents/drivers/test_utils_test.py
```diff
@@ -0,0 +1,58 @@
...
+      action=np.random.rand(batch_size, traj_len),
+      policy_info=(),
+      reward=np.random.rand(batch_size, traj_len),
+      discount=np.ones((batch_size, traj_len)),
+      step_type=np.zeros((batch_size, traj_len)),
+      next_step_type=next_step_type)
+
+    observer = driver_test_utils.NumEpisodesObserver()
+    observer(traj)
+    assert observer.num_episodes.read_value() == batch_size
```
egonina: you can use self.assertEqual here (https://www.tensorflow.org/api_docs/python/tf/test/TestCase#assertEquals).

egonina (commented):
Thanks for adding the test!

tf_agents/drivers/test_utils_test.py
leroidauphin: Thanks for the review!

PR #325, opened 2020-03-06
Set TFPyEnv's reward shape per wrapped env's time_step_spec.

mhe500:
This allows environments to have non-scalar rewards.
Relates to https://github.com/tensorflow/agents/pull/239.

### DIFF
--- a/tf_agents/environments/tf_py_environment.py
+++ b/tf_agents/environments/tf_py_environment.py
@@ -151,9 +151,12 @@ class TFPyEnvironment(tf_environment.TFEnvironment):
                                           action_spec,
                                           batch_size)
 
-    # Gather all the dtypes of the elements in time_step.
+    # Gather all the dtypes and shapes of the elements in time_step.
     self._time_step_dtypes = [
         s.dtype for s in tf.nest.flatten(self.time_step_spec())
+    ]
+    self._time_step_shapes = [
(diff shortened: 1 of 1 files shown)


tf_agents/environments/tf_py_environment.py
```diff
@@ -333,7 +336,7 @@ def _set_names_and_shapes(self, step_type, reward, discount,
     batch_shape = tf.TensorShape(batch_shape)
     if not tf.executing_eagerly():
       # Shapes are not required in eager mode.
-      reward.set_shape(batch_shape)
+      reward.set_shape(batch_shape.concatenate(self._time_step_shapes[1]))
```
ebrevdo: Why [1]?  And why calculate all flattened shapes if you only use one of them?

ebrevdo (changes requested):

tf_agents/environments/tf_py_environment.py
mhe500: Hi @ebrevdo. Thanks for the feedback.

I agree, it is not as clear as it should be and not necessary to cache all shapes.. I've updated to not cache shapes in `__init__` and to explicitly query `self.time_step_spec().reward.shape` rather than referring to it by its index position after flattening (`[1]`).

ebrevdo (approved):

Issue #357, opened 2020-04-30
PPOAgent._train increments the train_step_counter by num_epochs rather than 1


leroidauphin:
The docstring for the `TFAgent_train` method states that:
`This method *must* increment self.train_step_counter exactly once.`

The `PPOAgent._train` method increments ` self.train_step_counter` by `self._num_epochs`. This is because the counter in incremented (by one) by the optimizer `apply_gradients` method inside the loop which repeats `self._num_epochs` times: 
https://github.com/tensorflow/agents/blob/master/tf_agents/agents/ppo/ppo_agent.py#L777

summer-yue:
Thanks for the contribution! I responded in the PR, copying it here just in case:

"I have a high level question. We are actually moving away from using epoch_num inside of ppo_agent.train, and handle epochs with tf.Data pipelines outside of the agent. Given that, is there still added benefit in separating train_step_counter and global step here?"

summer-yue:
Resolved. See the PR for more detailed discussions.

PR #358, opened 2020-04-30
Use a separate counter variable for train_step_counter in PPOAgent

leroidauphin:
Separate the train_step_counter in the PPOAgent into two parts: a new counter variable and the global step. This allows the agent.train_step_counter to be incremented exactly once per call to _train, while maintaining the existing behaviour that the global step counts the number of optimisation steps (incremented by num_epochs per call to _train).

### DIFF
--- a/tf_agents/agents/ppo/ppo_agent.py
+++ b/tf_agents/agents/ppo/ppo_agent.py
@@ -238,7 +238,7 @@ class PPOAgent(tf_agent.TFAgent):
       debug_summaries: A bool to gather debug summaries.
       summarize_grads_and_vars: If true, gradient summaries will be written.
       train_step_counter: An optional counter to increment every time the train
-        op is run.  Defaults to the global_step.
+        op is run. Defaults to a new global variable created by this class.
       name: The name of this agent. All variables in this module will fall under
         that name. Defaults to the class name.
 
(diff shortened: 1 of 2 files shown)


leroidauphin:
Proposed solution for #357

tfboyd:
Looking for a reviewer.

summer-yue:
Thanks for the contribution! I have a high level question. We are actually moving away from using `epoch_num` inside of ppo_agent.train, and handle epochs with tf.Data pipelines outside of the agent. Given that, is there still added benefit in separating train_step_counter and global step here?

leroidauphin:
> Thanks for the contribution! I have a high level question. We are actually moving away from using `epoch_num` inside of ppo_agent.train, and handle epochs with tf.Data pipelines outside of the agent. Given that, is there still added benefit in separating train_step_counter and global step here?

Possibly not. The purpose of this PR was to make the `_train` method of the PPOAgent fulfil the requirement in the super class docstring that the `train_step_counter` is incremented exactly once. If moving to the `tf.Data` pipeline achieves that goal there is no need for this PR and it should be closed.
Is it possible to find out more about the plans to move to `tf.Data` pipelines?

summer-yue:
Sounds good! Thank you. We have actually already marked the .gather_all() method in the replay buffers deprecated, and ask all new users to move to .as_dataset(). I could add some comments in the PPO agents' documentations too, and potentially get rid of epoch_num soon (how fast this can be done would depend on how many existing projects are relying on this).

Thanks again for the contribution. Closing this for now.

PR #372, opened 2020-05-09
Fixed Broken Link in DQN Tutorial

DylanCope:
Replaced `https://github.com/tensorflow/agents/blob/master/tf_agents/docs/python/tf_agents/drivers.md` with `https://www.tensorflow.org/agents/api_docs/python/tf_agents/drivers` in the cell that defines the data collection for the replay buffer.

### DIFF
--- a/docs/tutorials/1_dqn_tutorial.ipynb
+++ b/docs/tutorials/1_dqn_tutorial.ipynb
@@ -1129,18 +1129,6 @@
    "language": "python",
    "name": "python3"
   },
-  "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
-   "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.6.10"
-  },
   "pycharm": {
    "stem_cell": {
     "cell_type": "raw",
@@ -1152,5 +1140,5 @@
   }
  },
  "nbformat": 4,
- "nbformat_minor": 1
+ "nbformat_minor": 0
 }
--- a/docs/tutorials/1_dqn_tutorial.ipynb
+++ b/docs/tutorials/1_dqn_tutorial.ipynb
@@ -1,1144 +1,1144 @@
 {
- "cells": [
-  {
-   "cell_type": "markdown",
-   "metadata": {
-    "colab_type": "text",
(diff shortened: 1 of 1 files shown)


DylanCope:
> @googlebot I signed it!

docs/tutorials/1_dqn_tutorial.ipynb
```diff
@@ -1,1144 +1,1156 @@
...
+   "private_outputs": true,
+   "provenance": [],
+   "toc_visible": true
+  },
+  "kernelspec": {
+   "display_name": "Python 3",
+   "language": "python",
+   "name": "python3"
+  },
+  "language_info": {
```
tfboyd: If you can remove from here down, through I think line 1142 I can just accept this with zero effort.  If not, I need to lookup if we really want this extra info.  Off the top of my head I don't know and it will take time to sort it out and see if it matches the other files and what to do. I likely to be overly transparent. I just don't know the possibly ramification of these additions.  Thank you for understanding and for the link fix.

tfboyd (changes requested):
Check out the comment and then we can merge this. Thanks.

tfboyd:
@DylanCope   Looks like this commit (last one too I just did not notice) changed \u003ctable into `<table>`.  I can make the change and glad you pointed it out. I like people to get the commit history in their github profile because it is fun.  Let me know. There is no rush. I want you to know I am cool fixing it and I am totally cool getting this PR fixed up. I am not awesome at colabs, I am not sure why the unicode got processed.  

Sorry I did not notice this the first time.  I was curious why the diff was showing "everything" changed vs. just your link fix.

DylanCope:
> @DylanCope Looks like this commit (last one too I just did not notice) changed \u003ctable into `<table>`. I can make the change and glad you pointed it out.

I had just noticed this too! I was just about to just copy across that portion of the notebook to fix it - not sure what effect it has.

I think it has changed the indentation on the file which is why it is showing up as everything having changed.

tfboyd:
I would just take the original and make the small change and push that.  The indention changing is kind of a pain and not desired.

DylanCope:
I went for a manual edit of the notebook file - looks much cleaner now

tfboyd:
Looks great.  I know this was a small change and I appreciate you making it perfect.  We have a small issue with our import tools, but I will get this merged ASAP.  You have my word, which is pretty good historically.

tfboyd (approved):

DylanCope:
Cheers, and no worries its good to get it right! :)

PR #380, opened 2020-05-19
Update uses of deprecated AffineScalar bijector.

### LINKED ISSUE #365: `AffineScalar` bijector is deprecated
`AffineScalar` bijector is deprecated; please use `tfb.Shift(loc)(tfb.Scale(...))` instead.

It is used in: 
tf_agents\distributions\utils.py:92: AffineScalar.__init__ (from tensorflow_probability.python.bijectors.affine_scalar)

leroidauphin:
The AffineScalar bijector is marked as deprecated. Uses of this bijector has been updated, and a test has been added.

fix #365 .


tfboyd:
This has been approved internally and should be auto merged in 24 hours or less I hope. Thanks for the fix.

PR #386, opened 2020-05-28
Use ==/!= to compare constant literals (str, bytes, int, float, tuple)

cclauss:
Avoid [SyntaxWarnings on Python >= 3.8](https://docs.python.org/3/whatsnew/3.8.html#porting-to-python-3-8).

% `python3.8`
```
>>> () is ()
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```

### DIFF
--- a/tf_agents/policies/tf_policy.py
+++ b/tf_agents/policies/tf_policy.py
@@ -239,7 +239,7 @@ class TFPolicy(tf.Module):
     return self._get_initial_state(batch_size)
 
   def _maybe_reset_state(self, time_step, policy_state):
-    if policy_state is ():  # pylint: disable=literal-comparison
+    if policy_state == ():
       return policy_state
 
     batch_size = tf.compat.dimension_value(time_step.discount.shape[0])


kyscg (approved):
This is great, thanks for your contribution!

midopooler (approved):

shivaylamba (approved):

tf_agents/policies/tf_policy.py
```diff
@@ -239,7 +239,7 @@ def get_initial_state(self, batch_size):
     return self._get_initial_state(batch_size)
 
   def _maybe_reset_state(self, time_step, policy_state):
-    if policy_state is ():  # pylint: disable=literal-comparison
+    if policy_state == ():
```
ebrevdo, line 242: this will cause failures if policy_state is a Tensor, and the output of policy_state == () will be a bool tensor.  we dont want that here since it's mean tto be a check on the type of policy_state.

feel free to add a comment to this effect.

Issue #398, opened 2020-06-09
Remove imports from `__future__`


leroidauphin:
The Python standard library includes a module called `__future__`, which contains function definitions which document (among other things) when breaking changes were introduced into the Python language. This module was of particular value around the Python 2/Python 3 boundary.

The TF-Agents project no longer supports Python 2, and includes code which is only valid in Python 3. Given that there is no longer official support for Python 2, it is now possible to remove most if not all of the import statements from the `__future__` module.

tfboyd:
This is totally logical. Unfortunately, we have a couple teams still using Python2 internally our internal build system works a bit differently than most and they cannot really pin to a specific "release" like an external user as a temp solution. My educated guess is we can remove __future__ in 2021.

tfboyd:
This has started to happen and all new code is PY3 only.  I am going to close this as we are not actively watching this issue and future will get removed over time.

PR #399, opened 2020-06-09
Remove imports from __future__.

leroidauphin:
These functions no longer need to be imported from `__future__` now that the TF-Agents project no longer supports Python 2.

c.f. #398

### DIFF
--- a/build_docs.py
+++ b/build_docs.py
@@ -23,10 +23,6 @@ python build_docs.py --output_dir=/path/to/output
 
 """
 
-from __future__ import absolute_import
-from __future__ import division
-from __future__ import print_function
-
 import os
 import sys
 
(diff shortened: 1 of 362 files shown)


leroidauphin:
It looks like there is an error from import/copybara. I don't know what the error is, but I would be keen to find out and happy to attempt to fix it.

tfboyd:
It may take a little while, but I will followup on this change. Since we dropped Python2, I think this is logical cleanup. 

Step 1: Do we want to do this?
Step 2: Do we want to do a PR or will this be easier to do internally?
  - fixing the conflict and then running the external and internal presubmit will provide the data needed to pick the right path.

tfboyd:
@leroidauphin I hope this was not too much work. We need to keep the __future__ due to some people still using python2 internally. I am updating the Issue as well.  My educated guess is hopefully we can remove __future__ in Q1 2021.  I wish I could say Q4 2020 but I have been doing this way too long to believe that. :-)   Thank you as usual for your work and sorry. This would be great if we could take it.

leroidauphin:
No worries :-) Thanks for the explanation!

PR #413, opened 2020-07-04
Small typo fix in the Policies tutorial

mmoffitt:


tfboyd:
Thanks, it should get merged in 24-48 hours. There is a quick internal review regardless of how small it is. I pushed it through step one and the review should be a 5 second LGTM. Saying 24-48 as sometimes people are not watching. I appreciate the typo or spelling fix.  :-)

tfboyd (approved):

PR #416, opened 2020-07-10
 Replace ddpg/critic_network.py mlp_layers with encoding networks.

rk37:
do kbanoop's TODO

### DIFF
--- a/tf_agents/agents/ddpg/critic_network.py
+++ b/tf_agents/agents/ddpg/critic_network.py
@@ -28,6 +28,9 @@ class CriticNetwork(network.Network):
 
   def __init__(self,
                input_tensor_spec,
+               preprocessing_layers=None,
+               preprocessing_combiner=None,
+               batch_squash=True,
                observation_conv_layer_params=None,
                observation_fc_layer_params=None,
                observation_dropout_layer_params=None,
@@ -45,6 +48,18 @@ class CriticNetwork(network.Network):
     Args:
       input_tensor_spec: A tuple of (observation, action) each a nest of
         `tensor_spec.TensorSpec` representing the inputs.
+      preprocessing_layers: (Optional.) A nest of `tf.keras.layers.Layer`
+        representing preprocessing for the different observations.
+        All of these layers must not be already built. For more details see
+        the documentation of `networks.EncodingNetwork`.
+      preprocessing_combiner: (Optional.) A keras layer that takes a flat list
+        of tensors and combines them. Good options include
+        `tf.keras.layers.Add` and `tf.keras.layers.Concatenate(axis=-1)`.
+        This layer must not be already built. For more details see
+        the documentation of `networks.EncodingNetwork`.
+      batch_squash: If True the outer_ranks of the observation are squashed into
+        the batch dimension. This allow encoding networks to be used with
+        observations with shape [BxTx...].
       observation_conv_layer_params: Optional list of convolution layer
         parameters for observations, where each item is a length-three tuple
         indicating (num_units, kernel_size, stride).
@@ -119,14 +134,18 @@ class CriticNetwork(network.Network):
       last_kernel_initializer = tf.keras.initializers.RandomUniform(
           minval=-0.003, maxval=0.003)
 
-    # TODO(kbanoop): Replace mlp_layers with encoding networks.
-    self._observation_layers = utils.mlp_layers(
-        observation_conv_layer_params,
-        observation_fc_layer_params,
-        observation_dropout_layer_params,
+    encoder = encoding_network.EncodingNetwork(
+        observation_spec,
+        preprocessing_layers=preprocessing_layers,
+        preprocessing_combiner=preprocessing_combiner,
+        conv_layer_params=observation_conv_layer_params,
+        fc_layer_params=observation_fc_layer_params,
+        dropout_layer_params=observation_dropout_layer_params,
         activation_fn=activation_fn,
         kernel_initializer=kernel_initializer,
-        name='observation_encoding')
+        batch_squash=batch_squash,
+        name='observation_encoding') 
+    self._encoder = encoder
 
     self._action_layers = utils.mlp_layers(
         None,
@@ -144,7 +163,7 @@ class CriticNetwork(network.Network):
         kernel_initializer=kernel_initializer,
         name='joint_mlp')
 
-    self._joint_layers.append(
+    self._joint_lay`ers.append(
         tf.keras.layers.Dense(
             1,
             activation=output_activation_fn,
@@ -154,10 +173,16 @@ class CriticNetwork(network.Network):
   def call(self, inputs, step_type=(), network_state=(), training=False):
     observations, actions = inputs
     del step_type  # unused.
-    observations = tf.cast(tf.nest.flatten(observations)[0], tf.float32)
-    for layer in self._observation_layers:
-      observations = layer(observations, training=training)
 
+    state, network_state = self._encoder(
+        observations,
+        step_type=step_type,
+        network_state=network_state,
+        training=training)
+    
+    # Treat as observations
+    observations = state
+    
     actions = tf.cast(tf.nest.flatten(actions)[0], tf.float32)
     for layer in self._action_layers:
       actions = layer(actions, training=training)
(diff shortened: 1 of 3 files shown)


tf_agents/agents/ddpg/critic_network.py
```diff
@@ -28,6 +29,9 @@ class CriticNetwork(network.Network):
 
   def __init__(self,
                input_tensor_spec,
+               preprocessing_layers=None,
+               preprocessing_combiner=None,
+               batch_squash=True,
```
ebrevdo, line 34: why do we need this argument publicly available?

ebrevdo (changes requested):
needs unit tests.

tf_agents/agents/ddpg/critic_network.py
rk37: This arguments will pass to encoding_network.EncodingNetwork create an EncodingNetwork  for combi observations and using custom observation preprocessing layers such as Attention.

ebrevdo: do you ever need to pass a value other than `batch_squash=True` to EncodingNetwork?

tfboyd: @rk37  Checking in to see if you are still moving forward?  Total cool if you are not, I try to keep things moving and clearly I failed a little here as it has been over a month.

rk37: > @rk37  Checking in to see if you are still moving forward?  Total cool if you are not, I try to keep things moving and clearly I failed a little here as it has been over a month. 

@tfboyd Sorry, I used the modified code to run my gym env, but I am not sure of its effectiveness in other envs.  I am using pytorch (stable baselines 3) to solve my problem recently. And I will do more tests and add unit test code later.

tfboyd: @rk37 no problem

momja:
Are there any updates on this? I tried using this for multi-observation SAC, but I've been getting errors when computing the critic loss.

tfboyd:
@momja  What error are you getting? It is nice to have a clean repro and it helps with motivation to get this stuff fixed ASAP. It also allows me to verify if it is fixed by the change or needs another change. If this commit resolves the error, maybe we can resurrect it ASAP. It was really close and then died.

ebrevdo:
+1 would love to see the error.  additionally we have a simplier approach to building Networks now that is not monolithic, using `tf_agents.networks.{Sequential, NestMap, NestFlatten}`.  You can find an example [here](https://github.com/tensorflow/agents/blob/89d189a8e75713b2549fda9e2cc8408e70758265/tf_agents/networks/nest_map_test.py#L72).  We plan to start using this code in more of our examples, but for now it may simplify your model building.

ebrevdo:
We just pushed [an example](https://github.com/tensorflow/agents/commit/b8be56b8a4cfd4085ac68d898389f8c59d4ae804#diff-c078e197628924bc0c7e40ed93dc37c4ae032397e9fff44bda2ee973b83d4097R129) for how to use SAC with Sequential.  We hope to update the other examples, and to do similar with DDPG soon.

momja:
I will be able to report on the error in the next few days, but right now I am testing out this new approach shared by @ebrevdo. I'm receiving some errors while trying to implement my own.

With the modified function:
```
def create_sequential_critic_net():
  value_layer = nest_map.NestMap(
      {
          "patch": patch_pre_layer,
          "color": color_pre_layer,
          "motion": motion_pre_layer
      }
  )
  value_layer = sequential.Sequential([
      value_layer,
      nest_map.NestFlatten(),
      tf.keras.layers.Concatenate(axis=-1)])

  action_layer = tf.keras.layers.Dense(10)

  def sum_value_and_action_out(value_and_action_out):
    value_out, action_out = value_and_action_out
    return tf.reshape(value_out + action_out, [-1])

  return sequential.Sequential([
      nest_map.NestMap((value_layer, action_layer)),
      tf.keras.layers.Lambda(sum_value_and_action_out)
  ])
```

Now if I want to  create variables: `create_sequential_critic_net().create_variables((observation_spec, action_spec))` I get the following error:
```
ValueError: `inputs` and `self.nested_layers` do not have matching structures:
  (DictWrapper({'color': ., 'motion': ., 'patch': .}), .)
vs.
  (., .)
Values:
  (DictWrapper({
'color': <tf.Tensor: shape=(1, 3), dtype=float32, numpy=array([[0.9768766 , 0.04171145, 0.3783009 ]], dtype=float32)>, 
'motion': <tf.Tensor: shape=(1, 3), dtype=float32, numpy=array([[ 5.0205655 , 11.804611  ,  0.68829167]], dtype=float32)>,
'patch': <tf.Tensor: shape=(1, 1, 1, 3), dtype=float32, numpy=array([[[[0.63228047, 0.84821343, 0.74741304]]]], dtype=float32)>}),
<tf.Tensor: shape=(1, 7), dtype=float32, numpy=
array([[-0.0612357 ,  0.27595967,  0.17574099, -0.2420119 ,  2.616015  ,
        -2.1489165 ,  0.69741297]], dtype=float32)>)
vs.
  (<tf_agents.networks.sequential.Sequential object at 0x7fbff42258e0>, <tf_agents.networks.sequential.Sequential object at 0x7fbff4230100>).
```
These structures look like they match to me. This is my current roadblock.

ebrevdo:
I think you're close!  I think your `value_layer` must instead be a dict passed to `NestMap`.  This is the easiest way to do it, and I'm going to be rethinking your use case to see if we can clean up `NestMap` so your version works properly the way you wrote it!

```python
value_layer_dict = {
          "patch": patch_pre_layer,
          "color": color_pre_layer,
          "motion": motion_pre_layer
      }

action_layer = tf.keras.layers.Dense(10)

def sum_value_and_action_out(value_and_action_out):
  value_out_dict, action_out = value_and_action_out
  value_out = tf.concatenate(tf.nest.flatten(value_out_dict), axis=-1)
  return tf.reshape(value_out + action_out, [-1])

  return sequential.Sequential([
      nest_map.NestMap((value_layer_dict, action_layer)),
      tf.keras.layers.Lambda(sum_value_and_action_out)
  ])
```

momja:
> 
> 
> I think you're close! I think your `value_layer` must instead be a dict passed to `NestMap`. This is the easiest way to do it, and I'm going to be rethinking your use case to see if we can clean up `NestMap` so your version works properly the way you wrote it!

You're right! this resolved that problem. I think in the future it would make sense to allow for this kind of structure. I'm still not able to test this out, as I'm still getting an error from `compute_variables` in the SACAgent.
```
AssertionError: Could not compute output KerasTensor(type_spec=TensorSpec(shape=(1, 4096), dtype=tf.float32, name=None), name='flatten/Reshape:0', description="created by layer 'flatten'")
  In call to configurable 'SacAgent' (<class 'tf_agents.agents.sac.sac_agent.SacAgent'>)
```

This I traced back to `patch_pre_layer`, which I provide in the value_out_dict. 

```
patch_pre_layer = tf.keras.models.Sequential(
    [
        tf.keras.layers.Conv2D(8, 3),
        tf.keras.layers.MaxPool2D(),
        tf.keras.layers.Flatten(),
    ]
)
```

However, if I use

```
patch_pre_layer = tf.keras.layers.Flatten()
```
I receive no errors. I am not sure if this is a problem with how SACAgent handles the network, or a problem with my construction.

ebrevdo:
Very interesting - I've never seen that before :(

You could try using `InnerReshape`, which we designed to work for these cases; does [this example](https://github.com/tensorflow/agents/blob/89d189a8e75713b2549fda9e2cc8408e70758265/tf_agents/networks/sequential_test.py#L107) help?.

momja:
I will try that! Have you tested this with different training batch sizes? When I pass an `experience_dataset_fn` to my learner with `batch_size > 1`, I receive the following error

```
ValueError: Input 0 of layer dense_3 is incompatible with the layer: expected axis -1 of input shape to have value 3681 but received input with shape (1, 942336)
```

I would guess this would have something to do with an unspecified batch_size when calling `create_variables` in SACAgent.

ebrevdo:
Is this when using keras Flatten?  keras will flatten inputs of type
[batch, time, ...] and inputs of type [batch, ...] into the same vector.
With tf-agents train() you expect [batch, time, ...] and at inference or in
policy you sometimes expect [batch, ...].  InnerReshape corrects that - it
allows both types of input..

tfboyd:
I have to give up. :-( We made great progress in 2020 as far as accepting PRs and doing so faster, but this didn't make it.

PR #417, opened 2020-07-11
tf 2.2.0 bandits tutorial fixed

daxiongshu:
The [original tutorial](https://github.com/tensorflow/agents/blob/v0.5.0/tf_agents/bandits/colabs/bandits_tutorial.ipynb) has several errors when running with tf 2.2.0. This PR fixes those errors.


daxiongshu:
@googlebot I signed it!

PR #418, opened 2020-07-11
fix bandits_tutorial with tf 2.2.0

daxiongshu:
The [bandits tutorial](https://github.com/tensorflow/agents/blob/master/tf_agents/bandits/colabs/bandits_tutorial.ipynb) has several errors with tf 2.2.0. This PR fixes them.


tfboyd:
Cool we are about to do a relaunch of bandits and I am going to add a nightly test for the colabs.  I will take a look.  They also should fix this to not need compat.v1 but #1 is have it working first.  I'll try to verify this ASAP and thank you.

tfboyd:
Looks good. I had to make a small change changing Base back to TFPolicy.  Base did not work I did not look into as I had a feeling the rest of your fix would resolve the problem and I could move forward with limited time spent.  I think the error was not an attribute or something.  I just rolled that part back and it was all good. The process is to review internally and it gets auto submitted.  I hope I can do that in the next 24 hours.  ON the plus side I need your fix or might new nightly tests with the bandits colab included will keep failing.  A nice incentive. :-)

tfboyd:
I want to say that this now works if you use tf-agents-nightly and will work when we release tf-agents for TF 2.3 which should be in a few weeks.  The change from Base to TFPolicy was done recently and fails if using TF agents 0.5.0 (current stable). The head is meant to work with tf-agents-nightly.  I also just made a cl/pr that adds  `pip install tf-agents` so the notebook will run as expected in colab. It should "land" in 72 hours.  Thanks again for the PR it kicked off additional work we should have been doing.

PR #424, opened 2020-07-19
Fix a typo in 5_replay_buffers_tutorial.ipynb

shuvoxcd01:
Original line: 

> Now that we know how to **created** a replay buffer, write items to it and read from it, we can use it to store trajectories during training of our agents. 

Fixed line: 

> Now that we know how to **create** a replay buffer, write items to it and read from it, we can use it to store trajectories during training of our agents.

### DIFF
--- a/docs/tutorials/5_replay_buffers_tutorial.ipynb
+++ b/docs/tutorials/5_replay_buffers_tutorial.ipynb
@@ -1,506 +1,518 @@
 {
-  "cells": [
-    {
-      "cell_type": "markdown",
-      "metadata": {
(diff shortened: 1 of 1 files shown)


shuvoxcd01:
@googlebot I signed it!

shuvoxcd01:
I made the change from Jupyter notebook. I am not 100% sure but it seems like the notebook first deleted and then rewritten the whole content. 
It was just a one word change but it looked like I changed the whole notebook - which is why I closed this pull request. I made another pull request with proper changes.

PR #434, opened 2020-07-25
Update 10_checkpointer_policysaver_tutorial.ipynb

shuvoxcd01:
When instantiating `DynamicStepDriver`, the `num_steps parameter` is given the value `collect_steps_per_iteration`. 
Hence, while training the agent, 
`  for _ in range(collect_steps_per_iteration):`
`  .... collect_driver.run()`
 lines will have the effect of collecting `collect_steps_per_iteration * collect_steps_per_iteration` steps of data where only `collect_steps_per_iteration` steps were intended.


docs/tutorials/10_checkpointer_policysaver_tutorial.ipynb
```diff
@@ -373,8 +373,7 @@
         "def train_one_iteration():\n",
         "\n",
         "  # Collect a few steps using collect_policy and save to the replay buffer.\n",
-        "  for _ in range(collect_steps_per_iteration):\n",
```
shuvoxcd01, line 376: `collect_steps_per_iteration` steps is taken (by DynamicStepDriver) for **each iteration** of this `for loop` (_since, at the time of instantiation, DynamicStepDriver got **num_steps=collect_steps_per_iteration**_).
Hence, at the end of this loop **collect_steps_per_iteration * collect_steps_per_iteration** number of steps will be taken.

egonina, line 376: Thank you for the PR and finding this issue! I think instead of removing the for loop we should change the range to be over `num_iterations`, i.e. we still want to run `num_iterations` iterations each collecting `collect_steps_per_iteration` steps. Does that make sense?

shuvoxcd01: I have a question. Do we want to run `num_iterations` iterations each collecting `collect_steps_per_iteration` steps within **one training iteration** ? [ These codes are within the function `def train_one_iteration()` ].
Now, if we do, then 'yes', it makes sense. Otherwise, no.

tfboyd: @egonina  Can you check out the question above?

egonina: I'm very sorry about missing your reply earlier! You are right, since this is running in one training iteration (num_iterations) iterates over train steps, this should be just `collect_driver.run()` like you have.

shuvoxcd01: Thank you @egonina  and @tfboyd.

tfboyd:
BOOM MERGED  :-)

PR #442, opened 2020-08-09
Reverse summation order in nested_kl_divergence() in PPO

### LINKED ISSUE #439: PPO bug on nested actions
Hi!

Sorry I wasn't able to put together a simple repro for this issue, but I pinned down where the error happens and will try to describe it here.

I have an actor distribution network which produces nested action with `action_spec` like this
```
{'alpha': BoundedTensorSpec(shape=(2,), dtype=tf.float32, minimum=-1, maximum=1),
 'beta': BoundedTensorSpec(shape=(2,), dtype=tf.float32, minimum=-1, maximum=1),
'theta': BoundedTensorSpec(shape=(1,), dtype=tf.float32, minimum=-1, maximum=1)} 
```
So it is a dictionary with 3 keys, and the values have different shapes (2,), (2,) and (1,). The problem in PPO arises in `ppo_utils.nested_kl_divergence()` when it calculated the sum of KL over all time steps. The line 
`summed_kl_divergences = tf.add_n(all_kl_divergences)`
expects that all dictionary values in this nested action are of the same shape, and it crashes in my example with the following error:
```
ValueError: Dimension 2 in both shapes must be equal, but are 2 and 1. Shapes are [10,12,2] and [10,12,1].
	From merging shape 1 with other shapes. for 'epoch_0/AddN_2' (op: 'AddN') with input shapes: [10,12,2], [10,12,2], [10,12,1].
```
Here 10 is batch size, 12 is number of steps in the episode, and it doesn't like the last dimension of the tensor, which is the dimension of my actions.

In the end this function `nested_kl_divergence()` does the `sum_reduce()` over all dimensions except the batch dimension anyways, so it seems that this error can be easily avoided by reversing the order of operations and doing the `reduce_sum()` first on  each element of the nested action and then adding all KL contributions together.

Could you fix this?

v-sivak:
nested_kl_divergence() in ppo_utils.py wasn't summing up the individual contributions to KL correctly, if the nested actions have different shapes. In this commit the order of operations is reversed: first the sum is done over time steps, which brings all KL contributions to the same shape, and then the sum over actions produces a total_kl. Fixes #439.

### DIFF
--- a/tf_agents/agents/ppo/ppo_utils.py
+++ b/tf_agents/agents/ppo/ppo_utils.py
@@ -126,14 +126,17 @@ def nested_kl_divergence(nested_from_distribution: types.NestedDistribution,
                         for from_dist, to_dist
                         in zip(flat_from_distribution, flat_to_distribution)]
 
+  all_kl_divergences_reduced = []
+  for kl_divergence in all_kl_divergences:
+    # Reduce_sum over non-batch dimensions.
+    reduce_dims = list(range(len(kl_divergence.shape)))
+    for dim in outer_dims:
+      reduce_dims.remove(dim)
+    all_kl_divergences_reduced.append(
+        tf.reduce_sum(input_tensor=kl_divergence, axis=reduce_dims))
+   
   # Sum the kl of the leaves.
-  summed_kl_divergences = tf.add_n(all_kl_divergences)
-
-  # Reduce_sum over non-batch dimensions.
-  reduce_dims = list(range(len(summed_kl_divergences.shape)))
-  for dim in outer_dims:
-    reduce_dims.remove(dim)
-  total_kl = tf.reduce_sum(input_tensor=summed_kl_divergences, axis=reduce_dims)
+  total_kl = tf.add_n(all_kl_divergences_reduced)
 
   return total_kl
 
--- a/tf_agents/agents/ppo/ppo_utils_test.py
+++ b/tf_agents/agents/ppo/ppo_utils_test.py
@@ -77,6 +77,22 @@ class PPOUtilsTest(parameterized.TestCase, tf.test.TestCase):
     kl_divergence_ = self.evaluate(kl_divergence)
     self.assertAllClose(expected_kl_divergence, kl_divergence_)
 
+    # test for distributions with different shapes
+    one_reshaped = tf.constant([[1.0]] * 3, dtype=tf.float32)
+    dist_neg_one_reshaped = tfp.distributions.Normal(
+        loc=-one_reshaped, scale=one_reshaped)
+    dist_one_reshaped = tfp.distributions.Normal(
+        loc=one_reshaped, scale=one_reshaped)
+
+    nested_dist1 = [dist_zero, [dist_neg_one_reshaped, dist_one]]
+    nested_dist2 = [dist_one, [dist_one_reshaped, dist_zero]]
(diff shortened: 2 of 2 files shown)


v-sivak:
@googlebot I signed it!

tf_agents/agents/ppo/ppo_utils.py
```diff
@@ -126,14 +126,17 @@ def nested_kl_divergence(nested_from_distribution: types.NestedDistribution,
...
+   
   # Sum the kl of the leaves.
-  summed_kl_divergences = tf.add_n(all_kl_divergences)
-
-  # Reduce_sum over non-batch dimensions.
-  reduce_dims = list(range(len(summed_kl_divergences.shape)))
-  for dim in outer_dims:
-    reduce_dims.remove(dim)
-  total_kl = tf.reduce_sum(input_tensor=summed_kl_divergences, axis=reduce_dims)
+  total_kl = tf.add_n(all_kl_divergences_reduced)
```
summer-yue, line 140: Great find, and thanks for the contribution! Could you please add a unit test for this to make sure it works? Can probably copy most of the existing `test_nested_kl_divergence` and change the action spec.

tfboyd:
Sorry for the delay.  I was out. I will try to nudge someone into taking a look on Monday. We **were** doing so well on reviewing PRs.

ebrevdo:
@summer-yue ptal?

summer-yue (approved):
This PR should fix the bug which assumed the shapes of components in the action to be the same. There's probably faster ways to do this but no bug is better than bug :) Thanks for the contribution!

tf_agents/agents/ppo/ppo_utils.py
v-sivak: Thanks! I agree with your bug philosophy. Added a unit test.

v-sivak:
By the way, unrelated to this PR, but I've written a `FlattenObservationWrapper` for tf policies, because a similar wrapper from `environments/wrappers` only works with py policies. I could push that too if you think it's useful (currently your `environments/tf_wrappers` is almost empty).

tfboyd:
@v-sivak  Trying to remember to ask the team today about your FlattenObservationWrapper and get back to you after.

tfboyd:
@ebrevdo  Will give you some info on the  `FlattenObservationWrapper` we had a great conversation and want to share the direction with you.

v-sivak:
Hey @ebrevdo any news here?

ebrevdo:
Apologies for not responding sooner.  Summer's approval LGTM; we should merge this.

Separately about `FlattenObserverWrapper`, we should try to avoid making envs that lose information by flattening.  Instead, we are working on mmore flexible network architectures that will allow you to process non-flattened data more easily.  A good example is [here](https://github.com/tensorflow/agents/blob/367abe3596d3efde92857d43f8d9642895383e1b/tf_agents/networks/nest_map_test.py#L82).  In particular, we've added a new `Sequential`, `NestMap`, and `NestFlatten` networks, which allow you to combine, build towers and flatten early on in your Network.

We hope this new approach to building networks can replace the old monolithic networks (e.g. QNetwork for DQN) which try to do everything in one place and aren't very flexible.

We're working on extending support for this to our actor networks, which have the additional complexity that they must emit `tfp.Distribution` objects.  In the meantime, you can use the new `Sequential` network for DQN, Behavioral Cloning, and any critic network.

PR #452, opened 2020-08-20
WIP: adding observation_and_action_constraint_splitters to PPO

niklasnolte:
This PR aims to enable to use observation_and_action_constraint splitters in the PPO agent.

It is WIP for two reasons:
1. the newly introduced code branches are not yet unit-tested
2. The unit test in ppo_policy_test seems to suggest that a ppo_policy already works fine with masking, because the one can simply wrap the networks for masking.
However, I am unsure how the observation normalization is supposed to work in that case.
Can someone educate me on that?

### DIFF
--- a/tf_agents/agents/ppo/ppo_agent.py
+++ b/tf_agents/agents/ppo/ppo_agent.py
@@ -139,6 +139,7 @@ class PPOAgent(tf_agent.TFAgent):
       debug_summaries: bool = False,
       summarize_grads_and_vars: bool = False,
       train_step_counter: Optional[tf.Variable] = None,
+      observation_and_action_constraint_splitter: types.Splitter = None,
       name: Optional[Text] = None):
     """Creates a PPO Agent.
 
@@ -320,6 +321,8 @@ class PPOAgent(tf_agent.TFAgent):
           'Only tf.keras.optimizers.Optimiers are well supported, got a '
           'non-TF2 optimizer: %s', self._optimizer)
 
+    self._observation_and_action_constraint_splitter = observation_and_action_constraint_splitter
+
     self._initial_adaptive_kl_beta = initial_adaptive_kl_beta
     if initial_adaptive_kl_beta > 0.0:
       # TODO(kbanoop): Rename create_variable.
@@ -335,9 +338,13 @@ class PPOAgent(tf_agent.TFAgent):
 
     self._observation_normalizer = None
     if normalize_observations:
+      observation = time_step_spec.observation
+      if self._observation_and_action_constraint_splitter:
+        observation, _ = self._observation_and_action_constraint_splitter(observation)
+
       self._observation_normalizer = (
           tensor_normalizer.StreamingTensorNormalizer(
-              time_step_spec.observation, scope='normalize_observations'))
+              observation, scope='normalize_observations'))
 
     self._advantage_normalizer = tensor_normalizer.StreamingTensorNormalizer(
         tensor_spec.TensorSpec([], tf.float32), scope='normalize_advantages')
@@ -349,6 +356,7 @@ class PPOAgent(tf_agent.TFAgent):
             actor_network=actor_net,
             value_network=value_net,
             observation_normalizer=self._observation_normalizer,
+            observation_and_action_constraint_splitter=self._observation_and_action_constraint_splitter,
             clip=False,
             collect=False))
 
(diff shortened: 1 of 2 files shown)


niklasnolte:
@googlebot I signed it!

ormandi:
Thank you for using and contributing into TF-Agents.

I think `PPOAgent` and the action masking (where the mask is provided in the observation) can be done without modifying the PPO agent using the `MaskSplitterNetwork` from (https://github.com/tensorflow/agents/blob/master/tf_agents/networks/mask_splitter_network.py).

Unfortunately we do not yet have an open sourced specific example to show how exactly this can be done (sorry for that), but here is a quick example to show how this can be done:

```python
from tf_agents.networks import mask_splitter_network
...

# Assumed `observation_spec` is the following dict:
# {
#   'observation': original_observation_spec,  # This is spec of the observation fed into the network.
#   'mask': action_mask_spec,                          # The spec of the action mask used in the top of the network.
# }
def observation_and_action_constraint_splitter(observation_and_mask):
  return observation_and_mask['observation'], observation_and_mask['mask']

actor_net = mask_splitter_network.MaskSplitterNetwork(
    observation_and_action_constraint_splitter,
    actor_distribution_network.ActorDistributionNetwork(
        observation_spec['observation'],
        action_tensor_spec,
        ...),
    passthrough_mask=True)

  value_net = mask_splitter_network.MaskSplitterNetwork(
      observation_and_action_constraint_splitter,
      value_network.ValueNetwork(
          observation_spec['observation'],
          ...)
      )

ppo = ppo_agent.PPOAgent(
      time_step_tensor_spec,
      action_tensor_spec,
      actor_net=actor_net,
      value_net=value_net,
     ...)
```

niklasnolte:
thanks @ormandi,
I actually tried this approach, found it in the ppo_policy_test (which to me served as good enough example). 
However, can you educate me on how observation normalization works in that case?
Unless I let the ppo_agent know about a splitter function, how can it take the observation out of its input to apply the normalizer on? (see the second bullet point in my initial comment)

niklasnolte:
thanks @ormandi,
I actually tried this approach, found it in the ppo_policy_test (which to me served as good enough example). 
However, can you educate me on how observation normalization works in that case?
Unless I let the ppo_agent know about a splitter function, how can it take the observation out of its input to apply the normalizer on? (see the second bullet point in my initial comment)

niklasnolte:
thanks @ormandi,
I actually tried this approach, found it in the ppo_policy_test (which to me served as good enough example). 
However, can you educate me on how observation normalization works in that case?
Unless I let the ppo_agent know about a splitter function, how can it take the observation out of its input to apply the normalizer on? (see the second bullet point in my initial comment)

ormandi:
Thank you, Niklas. You're right about these two things (observation normalization and splitter) seems incompatible at the moment. However, I'd make them compatible differently: How about passing the normalizer instance into the agent (and leaving it `None` (default) if no normalization required)? In this way we could create the proper normalizer outside of the agent which is compatible with the splitter. At the same time we could decrease the functionality included in the agent as opposed to increase it.

I was wondering if this is something that you could work with?

niklasnolte:
Alright. Do I understand correctly that keeping the agents lightweight is a general guideline for this project?

So IIUC you want to give the StreamingTensorNormalizer the possiblity to apply a splitter itself?
If so, do you envision it putting the split structure back together afterwards?
  * If so, how can this be done generically?
  * If not, do you think it is a good idea that a normalizer returns a structure with a different specification?
If not, I don't understand how you want to do it

I hope my nested ifs make sense, read them you would read python

niklasnolte:
Alright. Do I understand correctly that keeping the agents lightweight is a general guideline for this project?

So IIUC you want to give the StreamingTensorNormalizer the possiblity to apply a splitter itself?
If so, do you envision it putting the split structure back together afterwards?  
If you want to merge back the structures, how can this be done generically?  
If not, do you think it is a good idea that a normalizer returns a structure with a different specification?   

If on the other hand you don't want to give the splitter possibility to the normalizer, I am unsure how you want to have it done.

niklasnolte:
@ormandi, how shall we proceed here?

tfboyd:
This just kind of dropped off.  It has been a year.  This happens from time to time.  Closing to clean things up. Sorry it did not come to a conclusion.

PR #478, opened 2020-10-02
Fix typo: therafter -> thereafter

Bi0T1N:


Bi0T1N:
@googlebot I signed it!

Issue #281, opened 2020-01-06
dqn_tutorial - initial_collect_steps is defined to be 1000 but not used anywhere...


guyk1971:
The call to collect_data in Data Collection (section 1.8) should have been:
`collect_data(train_env, random_policy, replay_buffer, initial_collect_steps)` 
right ?

ebrevdo:
Thanks for catching this.  @adammichaelwood could you send a fix?

i guess we also need to set `initial_collect_steps` to `100` so the tutorial stays consistent.

guyk1971:
Sure, I'll send a fix. Pull request, right ?
Thanks

tfboyd:
fixed

PR #479, opened 2020-10-02
Fix for #281 unused variable in DQN tutorial

Denbergvanthijs:
I have noticed `initial_collect_steps` wasn't used in the DQN tutorial. Issue #281 has also addressed this.

Kind regards,
Thijs

### DIFF
--- a/docs/tutorials/1_dqn_tutorial.ipynb
+++ b/docs/tutorials/1_dqn_tutorial.ipynb
@@ -218,7 +218,7 @@
       "source": [
         "num_iterations = 20000 # @param {type:\"integer\"}\n",
         "\n",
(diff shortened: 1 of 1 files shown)


Denbergvanthijs:
@googlebot I signed it!

tfboyd:
Nice, it is so easy to miss this stuff and the story as to why gets lost.

Denbergvanthijs:
Thank you for merging my first ever commit on a big public repo!

tfboyd:
@Denbergvanthijs  I can tell you that I 100% understand your excitement. It is also a big reason we (tf-agents) try to merge the PRs. Also this is a nice catch if I had upgraded my testing script to say manipulate the number of steps thinking that variable was setting the value I would have been annoyed. 

Thank you for taking the time to build a PR and for reading the tutorial.

PR #485, opened 2020-10-06
Make use of existing function instead of for-loop in DQN tutorial

Denbergvanthijs:
Seems like a waste to not use `collect_data`.


yochliu8 (approved):

tfboyd:
Nice use of the API. I did a quick check and confirmed the obvious that it works the same. Sent for approval hope it gets merged today or with in 24-48 hours.

summer-yue (approved):
Looks good!

PR #501, opened 2020-10-24
Fix never ending recursion in Environments guide notebook

FlorisHoogenboom:
The "example" implementation given for the abstract method in the notebook is not an accurate description of what the implementation of the `_step` method of a `PyEnvironment` should do. Also, the current example would lead to a never ending recursion.


tfboyd:
@egonina  When you get time can you take a quick look and then we can move this forward?  Thanks.

egonina:
<p>Good catch, LGTM</p>

PR #524, opened 2020-11-27
Update README.md

lc0:
Fix tutorial link for bandits


tfboyd:
Sorry about this and thank you for finding it. I had to make a change to our import/export process. I am leaving this open until it is submitted and verified.

tfboyd:
Fixed.  Thanks and sorry this too so much back and forth. I had to change it internally because we have a script that exports the README and swaps internal google links with github links to the files. I would have much rather just taken your PR.

PR #549, opened 2021-02-03
Change from using gather_all to as_dataset in PPO example.

FlorisHoogenboom:
This PR intends to remove the use deprecated method `gather_all` in favor of using `as_dataset` in the PPO example. 

I have to admit that I am not 100% sure that this is intended to be used this way since it is nowhere documented explicitly, but this seems logical and does not hurt training performance it seems. I'm curious to hear any feedback on this.

### DIFF
--- a/tf_agents/agents/ppo/examples/v2/train_eval_clip_agent.py
+++ b/tf_agents/agents/ppo/examples/v2/train_eval_clip_agent.py
@@ -222,13 +222,13 @@ def train_eval(
         num_episodes=collect_episodes_per_iteration)
 
     def train_step():
-        ds = replay_buffer.as_dataset(
-            sample_batch_size=num_parallel_environments,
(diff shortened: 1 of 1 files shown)


egonina:
Thanks for the PR! What performance evaluation have you done with this? My concern is that if the buffer size is too large, one big batch may not fit in memory, as you are doing now, so we may need to iterate on batches. CC @summer-yue who most is familiar with this benchmark

oars:
Thanks for the contribution! Unfortunately we will be replacing this example with the one in the experimental directory shortly so I don't think we will be triple checking this setup. The replay buffer is also being replaced to rely on Reverb so there is not much incentive to go in this direction.

You can look at the experimental dir for  the newer example
https://github.com/tensorflow/agents/tree/master/tf_agents/experimental/examples/ppo

Issue #544, opened 2021-01-29
Feature Request: Support Duel DQN 


king821221:
A feature request proposal to add support of Duel DQN, as suggested in [paper](https://arxiv.org/pdf/1511.06581.pdf) [Dueling Network Architectures for Deep Reinforcement Learning] , which is described as:
"
The main benefit of this factoring is to generalize learning across actions without imposing any
change to the underlying reinforcement learning
algorithm. Our results show that this architecture leads to better policy evaluation in the presence of many similar-valued actions.
"

I have created a  network class to support the Duel DQN architecture, which could be plugged into the existing DQNAgent.
The pull request could be referred [here](https://github.com/tensorflow/agents/pull/551)

A brief summary of the proposed change:

Add a new class DuelQNetwork:  in new code file: tf_agents/networks/duel_q_network.py

self.encoder is the shared part to convert observation to common state tensor

self.a_encode_layer, self.a_value_layer is to project the common state tensor to action space in advantage branch
self.v_encode_layer, self.v_value_layer is to project the common state tensor to single scalar state value in state branch

apply the sum between the state value and the mean-adjusted (advantage - MEAN(advantage)) to get the Q-value

update train_eval.py to support train and eva via QNetwork and DuelQNetwork repsectively.

Following the experiment in the original DuelDQN paper, the network both uses 2 fully connected layers

The QNetwork uses fully connected layers (100, 100)
The DuelQNetwork uses fully connected layers (100) for the shared encoding part, and uses (50) for the advantage and state branch repsectively.

Their summaries could be referred to in the attached images. 
Q-network
![dqn_cart_pole_v0](https://user-images.githubusercontent.com/1598436/106230164-dadc8b80-6229-11eb-87b2-ec65d1d114ae.png)

Duel Q-Network
![duel_dqn_cart_pole_v0](https://user-images.githubusercontent.com/1598436/106230170-ddd77c00-6229-11eb-9f81-b3d21b00e4bb.png)

A corresponding pytorch [implementation](https://github.com/p-christ/Deep-Reinforcement-Learning-Algorithms-with-PyTorch/blob/master/agents/DQN_agents/Dueling_DDQN.py)

Expect receiving your feedback and suggestion upon the code and the experiement result.Thanks

ebrevdo:
Between this and other DQN advances, I think it's better to focus time on implementing and evaluating [IQN](http://proceedings.mlr.press/v80/dabney18a.html) and Distributional variants of CQL-DQN.

PR #551, opened 2021-02-07
Feature request proposal: support Duel_DQN

king821221:
related issue: [544](https://github.com/tensorflow/agents/issues/544)


tf_agents/agents/dqn/examples/v2/train_eval.py
```diff
@@ -87,6 +89,9 @@ def train_eval(
     train_sequence_length=1,
     # Params for QNetwork
     fc_layer_params=(100,),
+    a_fc_layer_params=(50,),
+    v_fc_layer_params=(50,),
+    network_type='duel_q_network',
```
egonina: let's not make this the default

tf_agents/agents/dqn/examples/v2/train_eval.py
```diff
@@ -160,7 +165,20 @@ def train_eval(
           output_fc_layer_params,
           num_actions)
     else:
-      q_net = create_feedforward_network(fc_layer_params, num_actions)
+      observation_spec = tf_env.time_step_spec().observation
+      action_spec = tf_env.action_spec()
+      if network_type == 'q_network':
```
egonina: what's the rationale behind adding this back? We've moved from using QNetwork to using sequential.Sequential (with create_feedforward_network)

egonina (commented):
Thanks for the review! A couple high-level comments. CC @sguada for another look

tf_agents/agents/dqn/examples/v2/train_eval.py
```diff
@@ -87,6 +89,9 @@ def train_eval(
     train_sequence_length=1,
     # Params for QNetwork
```
sguada, line 91: Instead of passing all the different parameters for all the different types of q_networks, instead pass the q_network_fn(observation_spec, action_spec) that builds the appropriate network, ex. q_network, duel_q_network, create_recurrent_network or create_feedforward_network

tf_agents/agents/dqn/examples/v2/train_eval.py
```diff
@@ -160,7 +165,20 @@ def train_eval(
...
           num_actions)
     else:
-      q_net = create_feedforward_network(fc_layer_params, num_actions)
+      observation_spec = tf_env.time_step_spec().observation
+      action_spec = tf_env.action_spec()
+      if network_type == 'q_network':
+        q_net = create_q_network(observation_spec,
+                                 action_spec,
+                                 fc_layer_params)
+      elif network_type == 'duel_q_network':
```
sguada: Move this outside.

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: Thanks for the review. I will modify it

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: This acts as a side by side option comparison as the duel q network.  Could you please share with me why moving from using QNetwork to sequential?

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: Done

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: I agree. However, the q_network and duel_q_network functions takes different number of arguments. In my understanding, we could use network_fn(**kwargs)?

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: You mean add a separate train_eval file to support using Q network and duel Q network? So we do not need to modify the original train_eval file?

sguada: I meant as an auxiliary function within the same file but outside of the train_eval()

Ex:

```
def q_network(observation_spec, action_spec, fc_layer_params=(100,)):
....

def dual_q_network(observation_spec, action_spec, fc_layer_params=(100,), a_fc_layer_params=(50,),
    v_fc_layer_params=(50,)):
   ....

```

tf_agents/agents/dqn/examples/v2/train_eval.py
sguada: I suggest we consolidate the first 2 arguments across all the networks to be `observation_spec, action_spec` and then each one may have their specific extra kwargs.

tf_agents/agents/dqn/examples/v2/train_eval.py
egonina: I was referring to this method here: https://github.com/tensorflow/agents/blob/54aaac0033985f17ffbf829af4e1487c8f43fd68/tf_agents/agents/dqn/examples/v2/train_eval.py#L341

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: So you mean changing the prototype of QNetwork class to accept variable number of arguments? Please correct me if I am not right. If that is the case, that change has broader impact and I would suggest doing it in a separate code refactoring task.

king821221: I have consolidated the q_network fns to a function q_network_fn.: [ab7bee2](https://github.com/tensorflow/agents/pull/551/commits/ab7bee286ead30d7c5d4353627f9da90b6f0c316)

tf_agents/agents/dqn/examples/v2/train_eval.py
king821221: please see the commit : ab7bee2

ebrevdo:
Suggestions:
* Remove all the stale non-Sequential networks (dual DQN networks etc) in favor of the new style networks
* Move dual DQN agent, policy, and train_eval under experimental and split it off from core DQN work.

king821221:
> Suggestions:
> 
> * Remove all the stale non-Sequential networks (dual DQN networks etc) in favor of the new style networks
[Reply]:  OK, please help provide some examples of the new style networks so that I could correct it

> * Move dual DQN agent, policy, and train_eval under experimental and split it off from core DQN work.
[Reply]: OK. My original thought is that the dual DQN is an extension to the original DQN network so I place its stuff in the same location as the DQN work.  So is it a standard to place the new proposed features implementation firstly under experimental ? When and what needs to be done to take it out of the experimental ?

ebrevdo:
We move things out of experimental once two conditions are met:

1.  2 or more teams/organizations are actively using the algorithm
2.  Once the first condition is met, we write an internal nightly regression test using the actor learner API to ensure that performance does not regress. Most of the associated code is open source, except for the internal infrastructure required to run the test.

At that point we move things out of experimental.

king821221:
Thank you for the clarification. I will move it to experimental and
re-submit

ebrevdo ***@***.***> 于2021年4月18日周日 上午5:24写道：

sguada:
There is no progress in this PR, closing for now, feel free to re-open if progress is made.

PR #552, opened 2021-02-07
Check failed when TimeStep.is_last() returns array-like object

zhouker94:
Return type of the function `is_last()` in the class `TimeStep` is:

`Bool = Union[bool, np.bool, Tensor, Array]`

which lead to the exception by using `if time_step.is_last():` if the return is array-like object:

`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`


egonina:
Thanks for the PR! Could you add some tests for this in tf_agents/evironments/utils_test.py?

zhouker94:
Hi, @egonina thanks for your reply.

I have added a test case to test array-like reward_spec. Somehow the test case failed, so I checked the [time_step](https://github.com/tensorflow/agents/blob/master/tf_agents/trajectories/time_step.py) module and might find the cause of this issue.

Please correct me if I'm wrong:

The `restart`, `transition` and `termination` functions construct `step_type` by inferring the shape from reward_spec. However, the return of the function [time_step_spec](https://github.com/tensorflow/agents/blob/master/tf_agents/trajectories/time_step.py#L364) always returns `step_type` as a scalar. So the validation for array-like reward_spec would always be failed.

If my assumption is legit, the inconsistency causes the test failed and needs to be fixed. Maybe in another PR?

egonina:
Yeah, I think you are correct. I am not sure what the reason for this inconsistency is, i.e. why time_step_spec looks at the first entry in `observation_spec`. CC @kbanoop, can you please take a look?

kbanoop:
Thanks for the PR! 

The time_step returned from an environment is array_like if the environment is batched. And I believe the validate_environment function does not support batched environments yet (@oars can confirm in case I am wrong). So first I wanted to check if you are seeing this error because your environment is batched or because of any other reason?

If it's because of batched environments and you would like to add support for it in validate_environment(), you have to add a couple more things:
1. When creating the policy you have to pass environment.batch_size to the policy constructor as the outer_dims argument. Because the policy also have to produce batched actions.
2. Instead of doing: if np.all(time_step.is_last): episode_count += 1, you have to do episode_count += np.sum(time_step.last). Since in a batched environment, there are multiple environments with episodes ending at different times.

zhouker94:
> Thanks for the PR!
> 
> The time_step returned from an environment is array_like if the environment is batched. And I believe the validate_environment function does not support batched environments yet (@oars can confirm in case I am wrong). So first I wanted to check if you are seeing this error because your environment is batched or because of any other reason?
> 
> If it's because of batched environments and you would like to add support for it in validate_environment(), you have to add a couple more things:
> 
> 1. When creating the policy you have to pass environment.batch_size to the policy constructor as the outer_dims argument. Because the policy also have to produce batched actions.
> 2. Instead of doing: if np.all(time_step.is_last): episode_count += 1, you have to do episode_count += np.sum(time_step.last). Since in a batched environment, there are multiple environments with episodes ending at different times.

hi @kbanoop, I saw this error because I was trying to build a multi-agent environment, which could return multi-dimensional rewards -> `R = [ r1, r2, r3 ... r_i, ... ]`, so `r_i` is the reward of agent i. All the agents start & stop at the same time. So I don't think it is a batched environment.

After I read the code, seems the environment will be inferred as batched if the `reward_spec` is set as multi-dimensional. Does this mean `tf-agent` doesn't support my case?

kbanoop:
Thanks for mentioning the multi-agent set up, now I understand better. I think the bug is in the restart/transition/termination functions.

The correct behavior is that the step type should be a scalar [] (or [batch_size] for batched environments), but it shouldn't depend on the reward shape. 

Specs don't contain batch size anyway, so in the spec, step_type is always a scalar. So I believe the time_step_spec() is doing the right thing:
https://github.com/tensorflow/agents/blob/56b3228f515d04b896920ed04fff074716479e2e/tf_agents/trajectories/time_step.py#L397

For the restart, transition etc functions, it seems to be inferring the batch_size/outer_dims from reward.shape or reward.shape[0] which could be incorrect for vector rewards as you said. As a quick work around, could you try calling the restart/transition/termination functions in your environment by explicitly providing the batch_size/outer_dims argument ?

zhouker94:
> Thanks for mentioning the multi-agent set up, now I understand better. I think the bug is in the restart/transition/termination functions.
> 
> The correct behavior is that the step type should be a scalar [] (or [batch_size] for batched environments), but it shouldn't depend on the reward shape.
> 
> Specs don't contain batch size anyway, so in the spec, step_type is always a scalar. So I believe the time_step_spec() is doing the right thing:
> https://github.com/tensorflow/agents/blob/56b3228f515d04b896920ed04fff074716479e2e/tf_agents/trajectories/time_step.py#L397
> 
> For the restart, transition etc functions, it seems to be inferring the batch_size/outer_dims from reward.shape or reward.shape[0] which could be incorrect for vector rewards as you said. As a quick work around, could you try calling the restart/transition/termination functions in your environment by explicitly providing the batch_size/outer_dims argument ?

Yep that works for me. Here is my workaround:

- for `ts.transition` set `outer_dims=()`
- for `ts.termination` set `outer_dims=()`
- for `ts.restart` set `batch_size=None`

Should we refactor those arguments' usage to keep consistency? Seems `batch_size` is using `None` or an Integer to indicate the batch size. I was confused by those arguments.

Should I close this PR and create a new issue related to this bug?

kbanoop:
Glad it worked out, and sorry for the confusion.  

Yes it will be great if you can close this and file a separate issue. I think the fix would be to make those arguments consistent, and may be also force those arguments to be provided explicitly when the reward is not a scalar. But we have to check if this will break a lot of users.

egonina:
Thanks, @kbanoop!

PR #559, opened 2021-02-15
set proper delimiter for movielens data

pstanisl:
File `u.data` use tabulator as a delimiter, i.e., load_movielens_data does not load any data.


pstanisl:
no CLA

PR #561, opened 2021-02-16
Use proper delimiter for movielens dataset

pstanisl:
A file `u.data` uses tabular as a delimiter. The original implementation does not correctly read and parse the data.

### DIFF
--- a/tf_agents/bandits/environments/dataset_utilities.py
+++ b/tf_agents/bandits/environments/dataset_utilities.py
@@ -156,7 +156,7 @@ def load_movielens_data(data_file):
   with tf.io.gfile.GFile(data_file, 'r') as infile:
     # The file is a csv with rows containing:
     # user id | item id | rating | timestamp
-    reader = csv.reader(infile)
+    reader = csv.reader(infile, delimiter='\t')
     for row in reader:
       user_id, item_id, rating, _ = row
       ratings_matrix[int(user_id) - 1, int(item_id) - 1] = float(rating)


bartokg (commented):
With the data file version we are using this delimiter breaks the test. Please consider using a version of the data file where the explicit delimiter is not necessary. Thanks

pstanisl:
Thank you @bartokg for your reply, but to completely honest. I did not find any information about the dataset, except class comment in [MovieLensPyEnvironment](https://github.com/tensorflow/agents/blob/cc305ee9c2b3e8ebcb6d2996c9551029c2a670d6/tf_agents/bandits/environments/movielens_py_environment.py#L33), where the URL is specified as ([Kaggle - MovieLens 100K Dataset](https://www.kaggle.com/prajitdatta/movielens-100k-dataset)). 

If I downloaded the dataset and opened `u.data` file, I discovered that it uses a tabulator as a separator. 

The data structure is the same if the dataset is downloaded from [GroupLens - MovieLens](http://files.grouplens.org/datasets/movielens/ml-100k.zip).

Is there any other source of the dataset or more info about the dataset which do you use?

bartokg:
We might be using an earlier version. We might update the load function at some point, but until then please go ahead and do a replace(tab, comma) on your file. Thanks!

bartokg:
Please see this change: now you can specify the delimiter with a parameter.
https://github.com/tensorflow/agents/commit/873fa84e20f27b2ce637e5bdae100ba6d3bb36f8
Thanks for raising our attention to this issue!

PR #563, opened 2021-02-17
Improve documentation and use code highlighting

Bi0T1N:


tfboyd:
+2 added thanks for the wording fix.

PR #576, opened 2021-03-18
Update reinforce agent

pedrogasg:
I notice that the reinforce agent is the only one that still have the global_step
I tried and the performance are quite the same for the Adam in V1 and keras
> keras.optimizers.Adam
Screenshot from 2021-03-18 17-07-04
> compat.v1.train.AdamOptimizer
Screenshot from 2021-03-18 16-26-37

Please tell me if this is correct or if you keep the global _step for a reason
Thanks


pedrogasg:
@googlebot I signed it!

Issue #507, opened 2020-11-06
SAC for discrete action (GumbelSoftmax reparameterization trick)


Kang-SungKu:
Hello, I need to make SacAgent work with discrete action, so try to implement GumbelSoftmax parameterization trick by re-defining the relevant classes. However, the calculation of `agent.train(experience).loss` fails after a few iterations, so would like to have some comments to address the issue. I understand that SacAgent is designed to only work with continuous action, but any inputs will be really appreciated.

Let me provide the current situation one-by-one: 1) my configuration, 2) what I have done, and 3) the error.

**First**, here are the versions I am working on:
- OS: Windows 10 version 2004 (build 19041.572)
- IDE: Pycharm 2020.2.3 (build: 202.7660.27)
- Python 3.8
- tf_agents: 0.6.0
- tensorflow: 2.3.1
- tensorflow-probability: 0.11.1
- numpy: 1.18.5

Also, here are the action spec and observation spec of my environment. The environment is a simplified version of Tetris, where the observation spec is a 2D board (while the env spec is shown as 3D, `self.depth` is 1 at this point), and action involves simply dropping a block from the top of the board (in other words, the number of actions = `board_width * the number of available blocks`). Please note that I am using `observation_and_action_constraint_splitter` to apply an action mask, such that infeasible action is not allowed. That's the reason the observation spec contains two different fields: 'observation' and 'mask':
~~~
self._action_spec = array_spec.BoundedArraySpec(
    shape=(), dtype=np.int32, minimum=0, maximum=self.action_spec_max, name='action')

# State space (3D array): x_coord X y_coord X z_coord, range(num_moves+1)
# Calculate theoretically maximum moves per game to bound the state
self._observation_spec = {
    'observation': array_spec.BoundedArraySpec(
        shape=(self.width, self.depth, self.height), dtype=np.int32,
        minimum=0, maximum=np.floor(self.num_moves_max), name='observation'),
    'mask': array_spec.BoundedArraySpec(shape=(self.action_spec_max+1,),
        dtype=np.int32, minimum=0, maximum=1, name='mask')
}
~~~

**Second**, I re-defined several classes related to `SacAgent`, to allow discrete action. Here, I am providing the part of the classes I modified (I omitted the lines duplicated with the original class).
~~~
class SacAgentDiscrete(sac_agent.SacAgent):
    def __init__(self,
                 ...,
                 temperature: types.Float = 0.01,
[...]

PR #578, opened 2021-03-21
Modify SAC to also allow discrete actions

DBraun:
Modify SAC to also allow discrete actions, based on "[Discrete and Continuous Action Representation for Practical RL in Video Games](https://arxiv.org/abs/1912.11077v1)" (Delalleau, Peter, Alonso, & Logut 2019). In a forward pass, the projection networks for the discrete actions are run. During training, the results are soft-maxed, but otherwise they're turned into one-hot vectors. These discrete results are concatenated to the original state and this then goes through the other projection networks, resulting in continuous actions. Then both kinds of actions are merged and returned.

Anything that was using SAC previously should work exactly the same as before 🤞. However, the safety rails have been lifted because now it allows discrete actions, too. In a personal project I removed the 1-action limit on `ddpg/critic_network.py` but that's not in this PR.

Related: https://github.com/tensorflow/agents/issues/507

I also apologize if this totally fails some tests because I was having a hard time running unit tests on Windows. I'll come back to try again in that case.

### DIFF
--- a/tf_agents/agents/sac/sac_agent.py
+++ b/tf_agents/agents/sac/sac_agent.py
@@ -250,12 +250,9 @@ class SacAgent(tf_agent.TFAgent):
         self.data_context, squeeze_time_dim=(train_sequence_length == 2))
 
   def _check_action_spec(self, action_spec):
-    flat_action_spec = tf.nest.flatten(action_spec)
-    for spec in flat_action_spec:
-      if spec.dtype.is_integer:
-        raise NotImplementedError(
-            'SacAgent does not currently support discrete actions. '
-            'Action spec: {}'.format(action_spec))
+    # The original SAC implementation would throw an error here if there were discrete actions,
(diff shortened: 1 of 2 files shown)


DBraun:
@googlebot I signed it!

DBraun:
@googlebot I fixed it, maybe?

ebrevdo:
We're working to deprecate large monolithic networks like `ActorDistributionRNNNetwork` in favor of using `tfa.networks.Sequential()`, which should allow you to implement the same computation.  Would you instead like to add a unit test for SAC that implements this algorithm using a Sequential network?

DBraun:
Thank you both for your suggestions. It'll be on my todo-list. I've got to figure out this CLA thing too without broadcasting my email to the world. 🤦

tfboyd:
A long time has passed and it looks like we cannot get a CLA signed.

PR #579, opened 2021-03-21
Change method name to policy.action.

shuvoxcd01:
Change method name from policy.step to policy.action.


oars:
Thanks for the contribution! Would you mind rebasing your branch to get fast forwarding of the branch to work?

Issue #554, opened 2021-02-10
Tutorial 1: Train a Deep Q Network with TF-Agents and the use of a driver


jan-gebauer:
Hello,

I believe the tutorial 1_dqn_tutorial.ipynb has an unnecessary import of dynamic_step_driver. The module is not used at all, so the scripts runs just fine when commented out. Furthermore, I think it would be nice if the Data Collecting section explicitly talked about tutorial 4 and that this pattern is called a Driver. It is already somewhat alluding to it:

```
# This loop is so common in RL, that we provide standard implementations. 
# For more details see the drivers module.
# https://www.tensorflow.org/agents/api_docs/python/tf_agents/drivers
```

Since the tutorial on Drivers exists, it seems very handy to me to use it. Something like this:

```
# This loop is very common in RL, we call it a "Driver" and we provide standard implementations. 
# For more details see the tutorial 4 or the drivers module
# https://github.com/tensorflow/agents/blob/master/docs/tutorials/4_drivers_tutorial.ipynb
# https://www.tensorflow.org/agents/api_docs/python/tf_agents/drivers
```

It's a very minor issue but I think it's a good idea to make it simpler for learners. It would have helped me.

ebrevdo:
Thanks!  We'd love for you to submit a PR cleaning this up!  It makes a great beginner contribution.

PR #583, opened 2021-04-02
Remove unused import in DQN tutorial

Veluga:
As alluded to by @jan-polivka in issue #554, the import of `tf_agents.drivers.dynamic_step_driver` in the DQN tutorial is not needed.


Veluga:
@googlebot I signed it!

PR #590, opened 2021-04-17
Run !sudo apt-get update in reinforce tutorial

alexhiggins732:
Fixes reinforce tutorial dependencies installation errors in Google Collab.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,4 @@ __pycache__
 .idea/
 .eggs/
 *.egg-info/
-/.vs
+.vs/
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ __pycache__
 *.swp
 .vscode/
 .idea/
+/.vs
--- a/docs/tutorials/6_reinforce_tutorial.ipynb
+++ b/docs/tutorials/6_reinforce_tutorial.ipynb
@@ -109,6 +109,7 @@
       },
       "outputs": [],
       "source": [
+        "!sudo apt-get update\n",
         "!sudo apt-get install -y xvfb ffmpeg\n",
         "!pip install 'imageio==2.4.0'\n",
         "!pip install pyvirtualdisplay\n",


alexhiggins732:
@googlebot I signed it!

sguada:
That seems only applicable to specific users or cases, not sure if it should be included. Maybe add it as a comment instead.

alexhiggins732:
> That seems only applicable to specific users or cases, not sure if it should be included. Maybe add it as a comment instead.

Understanble. However, the issue is that the [Google Colab tutorial](https://colab.research.google.com/github/tensorflow/agents/blob/master/docs/tutorials/6_reinforce_tutorial.ipynb) will not run without running this command first. Not sure what if adding a comment is the way to go to address that and assumed updating the Python Notebook would be the route to take. Especially for new people exploring or for non-Linux users who might be confused by asking them to manually run a command if they receive an error.

Additionally, not odds are that even Linux users running their own Notebooks might not have their apt up to date so odds are one or more packages referenced is likely to be missing. Then come in the MOOCs and other online courses that might also be looking to use the notebook without modification.

I personally prefer notebooks that just run without need to figure out what additional commands need to be ran. I can see many users just abandoning if the first cell they run gives them an error without troubleshooting. Even experienced users might think "Well, this notebook is out of date. No use in continuing."

Additionally, the addition of the `.vs` directive to gitigore would be nice for us Visual Studio users.

ebrevdo:
seems OK to me; the next command just installs ffmpeg via apt-get anyway.

sguada:
Ok let's do it then.

alexhiggins732:
> Ok let's do it then.

I have since notice a few other notebooks have the same issue. If you would like, I can PR updates to those as well.

sguada:
Please do
Sergio

alexhiggins732:
Done. Git didn't pick up the additional changes in this PR (only the fixed merge conflict), so I opened a new one: https://github.com/tensorflow/agents/pull/598
> Please do Sergio

alexhiggins732:
Merged with https://github.com/tensorflow/agents/pull/598

PR #601, opened 2021-04-24
Update random_tf_environment_test.py

ngroves08:
Added test for array based reward spec


tfboyd:
Note to self: I suspect this would go in after the other PR.

PR #616, opened 2021-06-02
environments/gym_wrapper forward reward_range from GymEnv to PyEnviro…

cmarlin:
The GymEnv defines a reward_range. This CL forwards reward_range in GymWrapper. Usefull for some algorithms instead of -inf;+inf default range

### DIFF
--- a/tf_agents/environments/gym_wrapper.py
+++ b/tf_agents/environments/gym_wrapper.py
@@ -165,6 +165,8 @@ class GymWrapper(py_environment.PyEnvironment):
     self._action_spec = spec_from_gym_space(self._gym_env.action_space,
                                             spec_dtype_map, simplify_box_bounds,
                                             'action')
+    self._reward_spec = specs.BoundedArraySpec(
+        shape=(), dtype=np.float64, minimum=gym_env.reward_range[0], maximum=gym_env.reward_range[1], name='reward')
     self._flat_obs_spec = tf.nest.flatten(self._observation_spec)
     self._render_kwargs = render_kwargs or {}
     self._info = None
@@ -247,6 +249,9 @@ class GymWrapper(py_environment.PyEnvironment):
 
   def action_spec(self) -> types.NestedArraySpec:
     return self._action_spec
+
+  def reward_spec(self) -> types.NestedArraySpec:
+    return self._reward_spec
 
   def close(self) -> None:
     return self._gym_env.close()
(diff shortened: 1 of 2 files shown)


tf_agents/environments/gym_wrapper.py
```diff
@@ -165,6 +165,8 @@ def __init__(self,
     self._action_spec = spec_from_gym_space(self._gym_env.action_space,
                                             spec_dtype_map, simplify_box_bounds,
                                             'action')
+    self._reward_spec = specs.BoundedArraySpec(
+        shape=(), dtype=np.float64, minimum=gym_env.reward_range[0], maximum=gym_env.reward_range[1], name='reward')
```
sguada: do all gym_env provide a reward_range?
Fix code indentation issues and long lines..

cmarlin: Yes, this is part of the API since may 2016 https://github.com/openai/gym/commit/99f40a2fc12ddbc8981c1dd60cf36552e058904e

I've fixed the long lines, but I'm not sure about indentation issue

tf_agents/environments/gym_wrapper.py
```diff
@@ -165,6 +165,11 @@ def __init__(self,
     self._action_spec = spec_from_gym_space(self._gym_env.action_space,
                                             spec_dtype_map, simplify_box_bounds,
                                             'action')
+    self._reward_spec = specs.BoundedArraySpec(shape=(),
```
sguada, line 168: Is the gym_env.reward always np.float64 scalar? Or can it be float32 and/or a vector?

tf_agents/environments/gym_wrapper_test.py
```diff
@@ -436,6 +436,13 @@ def test_obs_dtype(self):
     time_step = env.reset()
     self.assertEqual(env.observation_spec().dtype, time_step.observation.dtype)
 
+  def test_reward_range(self):
+    cartpole_env = gym.spec('CartPole-v1').make()
+    cartpole_env.reward_range = (-3.0, 5.0)
```
sguada, line 441: Are there any gym environment that already has a reward_range?

sguada (changes requested):

tf_agents/environments/gym_wrapper_test.py
cmarlin: Checking the OpenAI depot, only Frozen Lake (https://github.com/openai/gym/search?q=reward_range)
It's easier for Muzero, as you can declare game environment range but after test on many environment I automatically update reward range with the maximum discovered

cmarlin:
The feature doesn't appear to be so usefull

PR #647, opened 2021-08-18
Fix broken formatting of document.

rainwoodman:


tfboyd:
+2 for the internal review.  Just needs to pass tests and come back around to commit.

tfboyd (approved):

PR #712, opened 2022-02-25
Switch tf.losses.mse with tf.math.squared_difference in BehavioralCloningAgent

egordon:
When training a BC Agent with a continuous action space, I received the following warning:

```
WARNING:absl:Need to use a loss function that computes losses per sample, ex: replace losses.mean_squared_error with tf.math.squared_difference. Invalid value passed for `per_example_loss`. Expected a tensor tensor with at least rank 1
```
Should just be a 1-line swap to fix, unless there's something I'm missing (though this seems minor enough to not need an issue/discussion first).


tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
```diff
@@ -256,7 +256,7 @@ def _continuous_loss_fn(self, experience, training: bool = False):
     else:
       bc_action = bc_output
 
-    losses = tf.nest.map_structure(tf.losses.mse, experience.action, bc_action)
+    losses = tf.nest.map_structure(tf.math.squared_difference, experience.action, bc_action)
```
ebrevdo: Can you calculate the mean here?  changing from mse to squared_difference means this loss is no longer normalized by  batch size, and means existing learning rate may no longer work the same waay.

egordon: Thanks! But I'm not sure that's necessary.

I think the return of this function needs to be per-example loss, and batch-size scaling is handled by [`common.aggregate_losses`](https://www.tensorflow.org/agents/api_docs/python/tf_agents/utils/common/aggregate_losses)

Which is called here right afterwards: https://github.com/tensorflow/agents/blob/e1d637f224c225491c7424e8c4b6fa21f334f36e/tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py#L295

The previous mean calculation was redundant.

ebrevdo: Ah you're partly right; the aggregation is not over batch but over *time*.  So you're replacing average loss over time w/ sum loss over time; and that means if you have an RNN nework and subsequences, if you change the subsequence lengths then the loss will change.  I still think we want the mean here.

ebrevdo: (err if you change the subsequence lengths, the loss will clearly change, but more to the point it will no longer be normalized by the length; which was the old behavior)

egordon: Wait, I'm sorry, I'm a tad confused.

Previously, this function returned a scalar `tf.losses.mse(per_example_loss) == sum(squared_diff(per_example_loss))/per_example_loss.shape[0]` (i.e. sum / # batches)

This is passed into `aggregate_losses`, which calls `tf.nn.compute_average`, which in turn returns [`sum(per_example_loss) / num_batches(per_example_loss)`](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/python/ops/nn_impl.py#L472)

This sum appears to be over the entire tensor, and dividing by the [batch dimension](https://github.com/tensorflow/tensorflow/blob/3f878cff5b698b82eea85db2b60d65a2e320850e/tensorflow/python/ops/nn_impl.py#L460), not the 2nd (time) dimension.

So I'm not quite understanding why the MSE calculation isn't redundant.

ebrevdo: MSE is not dividing by per_example_loss.shape[0].  It normalizes by the count of non-batch entries:

```python
 tf.losses.mse(tf.constant([[0.5, 0.5, 0.5]]), tf.constant([[1.0, 1.0, 1.0]]))
>>> numpy=array([0.25], dtype=float32)
tf.losses.mse(tf.constant(2 * [[0.5, 0.5, 0.5]]), tf.constant(2 * [[1.0, 1.0, 1.0]]))
>>> numpy=array([0.25, 0.25], dtype=float32)
```

egordon: Ah, there we go, my mistake, I figured out what was happening on my end to cause the warning.

`tf.losses.mse` was collapsing the batch dimension in my case because the base action was rank 0 (i.e. shape(action) = [B, ]).

I'm changing the PR to swap the local loss function to `squared_difference` if the single continuous action is rank 0. But keep the proper MSE otherwise.

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
```diff
@@ -286,6 +286,12 @@ def _setup_as_continuous(self, time_step_spec, action_spec, loss_fn):
     collect_policy = actor_policy.ActorPolicy(
         time_step_spec, action_spec, actor_network=self._cloning_network)
     policy = greedy_policy.GreedyPolicy(collect_policy)
+
+    # Ensure local MSE doesn't collapse batch dimension
+    self._loss_mse_fn = tf.losses.mse
+    if tf.nest.flatten(action_spec)[0].shape.rank == 0:
```
ebrevdo: self._loss_mse_fn = tf.nest.map_structure(lambda s: tf.math.squared_difference if s.shape.rank==0 else tf.losses.mse, action_spec

(keep lines w/in 80 char col limit)

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
```diff
@@ -256,7 +256,7 @@ def _continuous_loss_fn(self, experience, training: bool = False):
     else:
       bc_action = bc_output
 
-    losses = tf.nest.map_structure(tf.losses.mse, experience.action, bc_action)
+    losses = tf.nest.map_structure(self._loss_mse_fn, experience.action, bc_action)
```
ebrevdo: map_structure(lambda loss, a, b: loss(a,b), self._loss_mse_fn, experience.action, bc_action)

[make sure loss_mse_fn is applied correctly elsewhere in this code via map_structure]

egordon:
Made the recommended changes to be more general, and tested that it works for my use case, though noting that, right now, the `map_structure` won't make much of a difference, as the local MSE function will only be called when the `action` is a single continuous value (i.e. `len(action_spec) == 1`).

Though we may want to investigate if the current implementation now admits multiple continuous actions and remove the single continuous check in `__init__`

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
```diff
@@ -256,7 +256,8 @@ def _continuous_loss_fn(self, experience, training: bool = False):
     else:
       bc_action = bc_output
 
-    losses = tf.nest.map_structure(tf.losses.mse, experience.action, bc_action)
+    losses = tf.nest.map_structure(lambda loss, a, b: loss(a,b),
```
sguada, line 259: Add test with multiple actions.

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
```diff
@@ -286,6 +287,12 @@ def _setup_as_continuous(self, time_step_spec, action_spec, loss_fn):
     collect_policy = actor_policy.ActorPolicy(
         time_step_spec, action_spec, actor_network=self._cloning_network)
     policy = greedy_policy.GreedyPolicy(collect_policy)
+
+    # Ensure local MSE doesn't collapse batch dimension
+    self._loss_mse_fn = tf.nest.map_structure(lambda s:
```
sguada, line 292: Add 2 different tests, one with action.shape.rank = 0 and another with rank > 0

sguada (changes requested):

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
egordon: @sguada Do you want me to add more functionality to this commit? I can remove the check for single, scalar discrete actions.
Right now, Multiple Actions are not allowed with the built-in loss functions, and I think adding them can probably be a separate PR.

tf_agents/agents/behavioral_cloning/behavioral_cloning_agent.py
egordon: @sguada What type of test are you looking for?

I added tests for multi-rank continuous trajectories (tested ranks 0, 1, and 2) that mimic `testTrainWithNN()`.

tfboyd:
Closing this as part of cleanup. I don't see this getting to the finish after over a year. Not the best outcome; and just being honest. I am merging some of the other PRs where I can resolve the issues in the next day or so.

PR #722, opened 2022-03-15
Support TFEnvironment in Actor

Rossil2012:
When using Actor to wrap a collecting policy on a TFEnvironment, I came across the exception "ValueError("Actor doesn't support TFEnvironments yet.")" and I found it easy to solve as long as to use TFDriver with the same arguments as the PyDriver, which drives the policy on a PyEnvironment.

### DIFF
--- a/tf_agents/train/actor.py
+++ b/tf_agents/train/actor.py
@@ -122,7 +122,13 @@ class Actor(object):
           max_steps=steps_per_run,
           max_episodes=episodes_per_run)
     elif isinstance(env, tf_environment.TFEnvironment):
-      raise ValueError("Actor doesn't support TFEnvironments yet.")
+      self._driver = tf_driver.TFDriver(
+          env,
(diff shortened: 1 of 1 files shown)


Rossil2012:
@ebrevdo Could you help review this PR please?

tf_agents/train/actor.py
```diff
@@ -122,7 +122,13 @@ def __init__(self,
           max_steps=steps_per_run,
           max_episodes=episodes_per_run)
     elif isinstance(env, tf_environment.TFEnvironment):
-      raise ValueError("Actor doesn't support TFEnvironments yet.")
+      self._driver = tf_driver.TFDriver(
```
sguada, line 125: Observers may not work with TFEnvironments since those are typically batched. Can you add a test to make sure this change works?

sguada (changes requested):

PR #731, opened 2022-04-05
Bias Layer: Add support for regularizer and constraint

saisua:
Small patch to add the bias_regularizer and bias_constraint. It is needed to be like other layers in TF2.0.
They are initialized as None based on current master.

### DIFF
--- a/tf_agents/keras_layers/bias_layer.py
+++ b/tf_agents/keras_layers/bias_layer.py
@@ -37,7 +37,10 @@ class BiasLayer(tf.keras.layers.Layer):
         shape `(batch_size, input_dim)`.
   """
 
-  def __init__(self, bias_initializer='zeros', bias_regularizer=None, bias_constraint=None, **kwargs):
+  def __init__(self, bias_initializer='zeros', 
+                     bias_regularizer=None, 
+                     bias_constraint=None, 
+                     **kwargs):
     if 'input_shape' not in kwargs and 'input_dim' in kwargs:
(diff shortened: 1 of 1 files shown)


tf_agents/keras_layers/bias_layer.py
```diff
@@ -39,12 +37,14 @@ class BiasLayer(tf.keras.layers.Layer):
         shape `(batch_size, input_dim)`.
   """
 
-  def __init__(self, bias_initializer='zeros', **kwargs):
+  def __init__(self, bias_initializer='zeros', bias_regularizer=None, bias_constraint=None, **kwargs):
```
sguada: Can you reformat so it doesn't go over 80 characters and each argument use one line. Ex:
def __init__(self, 
                     bias_initializer='zeros',
                     bias_regularizer=None,
                     bias_constraint=None, 
                     **kwargs):

saisua:
I just did that. Thank you for your quick response.
As a side note, I have seen that there is a file called something like "bias layer test". I have not touched it but feel free to ask for anything.

sguada:
Please add 2 new test to make sure the new arguments work as expected.

sguada (changes requested):
Add tests for the new parameters.

tfboyd:
This has been about 1.5 years. I am making this closed.  We can always reopen.  I am merging PRs where with my knowledge can get the merged. This one is too big for me to handle unfortunately. Sometimes things rot on the vine; sad and a reality. 

Thank you for the effort and sorry this did not make it to the finish.

PR #749, opened 2022-06-11
Rename `tf_agents.policies.policy_saver.PolicySaver`

chunduriv:
Rename `tf_agents.policies.policy_saver.PolicySaver` to `tf_agents.policies.PolicySaver`


tfboyd:
I am 99% sure the original is correct.  The path is tf_agents.policies.policy_saver.PolicySaver.  

Just doing some final clean up.  If you even see this comment, I bet it is weird to see something from almost 1.5 years ago.  Thank you for taking the time to make a PR and using TF-Agents.

Issue #747, opened 2022-06-05
Error while importing tf agents in google colab


ghost:
`from __future__ import absolute_import
from __future__ import division
from __future__ import print_function

import abc
import tensorflow as tf
import numpy as np
import pandas as pd

from tf_agents.environments import py_environment
from tf_agents.environments import tf_environment
from tf_agents.environments import tf_py_environment
from tf_agents.environments import utils
from tf_agents.specs import array_spec
from tf_agents.environments import wrappers
from tf_agents.environments import suite_gym
from tf_agents.trajectories import time_step as ts

tf.random.set_seed(12)
tf.config.list_physical_devices('GPU')` 

`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[<ipython-input-5-068198677c33>](https://localhost:8080/#) in <module>()
      8 import pandas as pd
      9 
---> 10 from tf_agents.environments import py_environment
     11 from tf_agents.environments import tf_environment
     12 from tf_agents.environments import tf_py_environment

10 frames
[/usr/local/lib/python3.7/dist-packages/tensorflow/python/util/deprecation.py](https://localhost:8080/#) in new_func(*args, **kwargs)
    546                 'in a future version' if date is None else ('after %s' % date),
    547                 instructions)
--> 548       return func(*args, **kwargs)
    549 
    550     doc = _add_deprecated_arg_notice_to_docstring(

TypeError: function() got an unexpected keyword argument 'reduce_retracing'`

Cant understand what the problem is need help to solve the issue. I also tried reinstalling libraries still the error persists.

xivarri:
Hi, I reported this today too here https://github.com/tensorflow/agents/issues/746

ghost:
should i close it

sguada:
Make sure to run `!pip install tf_agents[reverb]`

ghost:
> Make sure to run `!pip install tf_agents[reverb]`

what was the real problem can you explain me?

melvinmt:
@ashishbhong82 I found some documentation about `reverb` here: https://www.tensorflow.org/agents/overview#stable

PR #761, opened 2022-07-13
Fix reduce_retracing error in tutorial 2 on colab

RacingTadpole:
When trying to run tutorial 2 in colab, I got the error

```
TypeError: function() got an unexpected keyword argument 'reduce_retracing'`
```

This PR implements the resolution suggested in https://github.com/tensorflow/agents/issues/747, which works for me.


tfboyd:
This will fix it.  I am working to get this accepted.  Should be 24 hours of less. I hope in a few.  Thank you for catching this.  It is weird the default colab env seems to be using an older TF.

Issue #772, opened 2022-09-08
Batching ReverbAddEpisodeObserver with variable length episodes


philstahlfeld:
It looks like [this bug](https://github.com/tensorflow/agents/blob/v0.13.0/tf_agents/replay_buffers/reverb_utils.py#L50) is already tracking the issue, but I would like to use ParallelPyEnvironment in conjunction with ReverbAddEpisodeObserver where my environment has variable length episodes. Is there a timeline for addressing this bug? If not, would it be something I could contribute?

philstahlfeld:
I was able to get around this by creating an observer that unbatches trajectories and delegates those trajectories to a collection of ReverbAddEpisodeObservers.
```py

from typing import Callable

from tf_agents.trajectories import trajectory as trajectory_lib
from tf_agents.utils import nest_utils

Observer = Callable[[trajectory_lib.Trajectory], None]


class UnbatchingObserver(object):
    """An observer that unbatches trajectories.

    The unbatched trajectories are delegated to observers that don't support
    batch dimensions (e.g. ReverbAddEpisodeObserver).

    Note that the batch size is not validated.
    """

    def __init__(self,
                 create_unbatched_observer_fn: Callable[[], Observer],
                 batch_size: int):
        self._unbatched_observers = [
            create_unbatched_observer_fn() for _ in range(batch_size)
        ]

    def __call__(self, batched_trajectory: trajectory_lib.Trajectory):
        unbatched_trajectories = nest_utils.unstack_nested_arrays(
            batched_trajectory)
        for obs, traj in zip(self._unbatched_observers, unbatched_trajectories):
            obs(traj)
```

Then to create I create the Actor observer as follows:
```py
    def rb_observer_fn():
        return reverb_utils.ReverbAddEpisodeObserver(
            reverb.Client(reverb_address),
            table_name=reverb_replay_buffer.DEFAULT_TABLE,
            max_sequence_length=max_sequence_length,
        )
    rb_observer = unbatching_observer.UnbatchingObserver(
        rb_observer_fn, batch_size=num_environments)
```

Please let me know if the UnbatchingObserver is something that others would find useful and I can contribute it.

sguada:
Thanks for your contribution, I think it would be great to get this as a PR with a simple example using ParallelEnv.
I would probably suggest renaming UnbatchingObserver to BatchedObserverUnbatching

As extra points it would be great to use multi-threads for running the individual observers.

```
class BatchedObserverUnbatching(object):
[...]

PR #774, opened 2022-09-22
Implement batched observer unbatching

philstahlfeld:
This is essentially an adapter that allows observers which don't support a batch dimension (looking at you, ReverbObservers) to be used in batch contexts.


philstahlfeld:
Closes #772

tf_agents/utils/batched_observer_unbatching.py
```diff
@@ -0,0 +1,51 @@
...
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from typing import Callable
+
+from tf_agents.trajectories import trajectory as trajectory_lib
+from tf_agents.utils import nest_utils
+
+Observer = Callable[[trajectory_lib.Trajectory], None]
```
guanhangwu: This is only used once. You don't have to create a variable for it.

tf_agents/utils/batched_observer_unbatching_example.py
```diff
@@ -0,0 +1,66 @@
+# coding=utf-8
```
guanhangwu: This file only serves the purpose of testing. You can put it in the batched_observer_unbatching_example_test.py.

philstahlfeld: Done.

tf_agents/utils/batched_observer_unbatching.py
philstahlfeld: I created that for readability since:
```py
Callable[[], Observer]
```
seems to me to read better than:
```py
Callable[[], Callable[[trajectory_lib.Trajectory], None]]
```
WDYT?

guanhangwu: Callable[[], Callable[[trajectory_lib.Trajectory], None]] is fine, so people don't have to check what Observer means to use this function. If you want to use this in multiple places or multiple files, it's better to create a variable for it. If you want to elaborate on this, you can mention it in the comments section of __init__

guanhangwu (commented):
Thanks for your quick response.

tf_agents/utils/batched_observer_unbatching.py
philstahlfeld: Done.

Issue #784, opened 2022-10-17
Incorrect formatting for time_step, trajectory when using tf.print


coreyleveen:
When passing a `TimeStep` or `Trajectory` to `tf.print`, the output presents the objects incorrectly in that some field values get printed in the wrong places.

The following code demonstrates this:

```python
import tensorflow as tf

from tf_agents.trajectories import time_step
from tf_agents.trajectories import trajectory

ts = time_step.TimeStep(
  tf.constant(0),     # step_type
  tf.constant(0.5),   # reward
  tf.constant(0.99),  # discount
  tf.constant([1, 1]) # observation
)

tf.print(ts)
#=> TimeStep(
#{'discount': 0, 'observation': 0.5, 'reward': 0.99, 'step_type': [1 1]})

observation, action, policy_info = (), (), ()
reward = tf.constant([2.0, 2.0, 2.0])
discount = tf.constant([0.99, 0.99, 0.99])
traj = trajectory.first(observation, action, policy_info, reward, discount)

tf.print(traj)
#=> Trajectory(
#{'action': (),
# 'discount': [0 0 0],
# 'next_step_type': [1 1 1],
# 'observation': (),
# 'policy_info': (),
# 'reward': [2 2 2],
# 'step_type': [0.99 0.99 0.99]})
```

Note the incorrect values in the output of the above code.
This occurs because both `TimeStep` and `Trajectory` use `pprint.pformat` in their respective `__repr__` method, which sorts by the dict keys when printing. However, `tf.print` still prints field values in the order of the namedtuple's `_fields`, which results in a mismatch of field names and the associated values.

PR #785, opened 2022-10-17
Fix time_step, trajectory format for tf.print

coreyleveen:
Addresses #784.
Note, this requires `python >= 3.8`, so should probably be put on ice until a release requires 3.8.

Open to suggestions, as this approach will cause fields of `TimeStep` and `Trajectory` to no longer be printed in sorted order.
`Transition` is not affected here because the values of its fields are not `Tensor` objects, rather `PolicyStep` and `TimeStep` objects.

### DIFF
--- a/tf_agents/trajectories/time_step.py
+++ b/tf_agents/trajectories/time_step.py
@@ -95,8 +95,11 @@ class TimeStep(
     return hash(tuple(tf.nest.flatten(self)))
 
   def __repr__(self):
-    return 'TimeStep(\n' + \
-      pprint.pformat(dict(self._asdict()), sort_dicts=False) + ')'
+    return (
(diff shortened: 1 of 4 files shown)


tfboyd:
We are on 3.9 required so maybe we can take this as part of a "final" (for now) push to clear out the PRs. I gave it a +2 and will see if I can get someone else to give it a look.  WOW we are almost a 1 year on this one.  Not the oldest thing I have cleaned up.

tfboyd:
We did it.  4 more to go.  :-). I wish I could explain how hard (needlessly) it is to get these merged. I am glad we made a push to get 5 PRs in that were crazy stale. No idea why; but I really wanted to get these in.

PR #796, opened 2022-11-19
Update actor/learner, distributed docs

coreyleveen:
- Updates actor/learner and distributed documentation to address some inaccuracies with API usage, and some related cleanup/clarification.


tfboyd:
Sorry this is about a year later.  I am doing one last clean up and this is nice. Thank you for taking the time.  I gave it a +2 internally and will try to remember to get someone to do the finally approval.  I cannot do it myself, it takes two people. 

We are continue to do releases and keep the stable version working. I am trying to pick up some old PRs; but we do not have a rotation anymore so it is best effort, e.g. quarters for someone to notice and review/merge. :-(. I am doing one last clean up.

PR #804, opened 2022-12-11
Documentation: Fix broken Deep Q Network with TFA build

8bitmp3:
- Upgrades TF to 2.11 (Colab uses 2.9.2), upgrades other dependencies (PiPy) just in case.

```
!pip install -U -qq 'imageio==2.4.0' pyvirtualdisplay tf-agents[reverb] pyglet tensorflow
```

Without this fix, you get:

```
RuntimeError                              Traceback (most recent call last)
RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xe
```

after attempting to import the necessary libraries.

@tfboyd @MarkDaoust PTAL

### DIFF
--- a/docs/tutorials/1_dqn_tutorial.ipynb
+++ b/docs/tutorials/1_dqn_tutorial.ipynb
@@ -114,10 +114,7 @@
       "source": [
         "!sudo apt-get update\n",
         "!sudo apt-get install -y xvfb ffmpeg freeglut3-dev\n",
-        "!pip install 'imageio==2.4.0'\n",
-        "!pip install pyvirtualdisplay\n",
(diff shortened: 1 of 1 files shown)


MarkDaoust:
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
RuntimeError: module compiled against API version 0x10 but this version of numpy is 0xe
```

Both versions (before and after) print that and continue running, so I'm not sure this changes anything. 

The fix Toby is working on is because of numpy version changes, so that may likely clear this warning.

8bitmp3:
> Both versions (before and after) print that and continue running, so I'm not sure this changes anything.

Thanks @MarkDaoust . Closing the PR.

PR #826, opened 2023-03-07
Upgrade numpy version in  1_dqn_tutorial.ipynb

kiransair:
Added a code cell to upgrade numpy version because while importing the libraries facing an error: `module compiled against API version 0x10 but this version of numpy is 0xf`. Upgrading the numpy version resolves the error.

### DIFF
--- a/docs/tutorials/1_dqn_tutorial.ipynb
+++ b/docs/tutorials/1_dqn_tutorial.ipynb
@@ -122,6 +122,14 @@
     },
     {
       "cell_type": "code",
+      "outputs": [],
+      "source": [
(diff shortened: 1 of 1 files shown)


tfboyd:
It should be good as-is. We also often need a specific version of numpy. At times the range is wide and others the range is minimal.  After doing this open source for 4-5 years with TensorFlow, TF-Agents, and others I am shocked that anything ever works at all.  I could write a short novel on the drama with Gym and some other lesser known packages that people count on that are not really maintained.  I also appreciate how hard it is to maintain these things.

Issue #585, opened 2021-04-05
validate_py_environment doesn't check that reset() is called on episode termination


AbhijeetKrishnan:
`PyEnvironment` objects assume that the `step()` method calls the `reset()` method if the episode has terminated. `validate_py_environment` only checks that the time_step_spec matches the returned time_step, and does not check if `reset()` is called during `step()` if the episode has ended.

This is difficult to debug when writing a custom environment. For example, my custom environment returns an error when `step()` is called after episode termination with a non-`None` action. I was trying to use `DynamicEpisodeDriver` to fill a replay buffer, and this error was repeatedly raised. Tracking down the issue with my environment not meeting the assumption regarding `reset()` was difficult.

This behavior can be seen by commenting out the following lines in the `CardGameEnv` class from the [TF-Agents Environments](https://www.tensorflow.org/agents/tutorials/2_environments_tutorial) tutorial.

```python
    if self._episode_ended:
        # The last action ended the episode. Ignore the current action and start
        # a new episode.
    return self.reset()
```

I would expect `validate_py_environment` to return an error, but it does not.

sguada:
Thanks for raising the issue. I can think a couple of solutions, one is to add a parameter `auto_reset` to the base `PyEnvironment` class that makes explicit the handling of auto-reset, although it will be false by default. And then make sure that `validate_py_environment` checks if that auto reset happens when is `True`.

Would you like to do a PR to fix this?

AbhijeetKrishnan:
That sounds like a clean way to implement it. I would be happy to submit a PR.

AbhijeetKrishnan:
I believe a better way to solve this would be to modify [`environments/py_environment.py:200`](https://github.com/tensorflow/agents/blob/master/tf_agents/environments/py_environment.py#L200) as follows -

```python
if self._current_time_step is None or self._current_time_step.is_last():
    return self.reset()
```

This would call `reset()` if the `_current_time_step` (which at this point in the function contains the *previous* time step) is of type `StepType.LAST`, indicating episode end. This change also seems to better reflect the assumptions made. Using `auto_reset` as a parameter seems to require making changes in a lot of other places. At least, one would have to check in `DynamicEpisodeDriver` that `auto_reset` is `True`.

[...]

PR #851, opened 2023-07-26
Added auto-reset handling to PyEnv Wrappers

sehejjain:
After #585 `handle_auto_reset` was added to the PyEnvironment class. This PR adds this as an init argument for the PyEnvironment wrappers. 

I had to use the PyEnvironment wrapper in my project and use this argument. Might be useful in other wrappers as well.


sehejjain:
@tfboyd Can you please review this?

